PR #3, opened 2024-02-18
Add "api/v1/metadata" endpoint.

leo-ship-it:
This endpoint is also required to integrate with DefiLlama, it returns all the metadata of the Taggr canister.


xqxpx (approved):
Super cool!

PR #4, opened 2024-02-18
Some suggestions from Grammarly

leo-ship-it:
Grammarly complained with some sentences in the whitepaper ;)

https://app.grammarly.com/

### DIFF
--- a/src/frontend/src/common.tsx
+++ b/src/frontend/src/common.tsx
@@ -1,15 +1,23 @@
 import * as React from "react";
 // @ts-ignore
 import DiffMatchPatch from "diff-match-patch";
-import { Clipboard, ClipboardCheck, Close, Flag, Menu, Share } from "./icons";
+import {
+    CarretDown,
+    Clipboard,
+    ClipboardCheck,
+    Close,
+    Flag,
+    Menu,
+    Share,
+} from "./icons";
 import { loadFile } from "./form";
-import { Post, PostId, Report, User, UserId } from "./types";
+import { Post, PostId, Report, User, UserFilter, UserId } from "./types";
 import { createRoot } from "react-dom/client";
 import { Principal } from "@dfinity/principal";
-import { IcrcAccount } from "@dfinity/ledger";
+import { IcrcAccount } from "@dfinity/ledger-icrc";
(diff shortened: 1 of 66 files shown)


src/frontend/src/common.tsx
```diff
@@ -974,6 +974,7 @@ export const icrcTransfer = async (
     } catch (e) {
         return "Transfer failed";
     }
+    return response;
```
taggrx: This is wrong, this variable is undefined here.

xqxpx (commented):
Thanks so much @leo-ship-it! This was long overdue!

Please fix the JS bug and we can merge.

src/frontend/src/common.tsx
leo-ship-it: My bad.

mechaquan:
Btw, on the pop up message when trying to transfer tokens it says you voted on an open proposal, right?  I think this should be changed as well.

I'm not sure where this message exists so I can't confirm the exact wording but I think it should be clarified.  Simple "Tokens cannot be transferred while there is an open proposal."

leo-ship-it:
@mechaquan Currently the error message you get is: `Transfer failed. One reason might be that you voted on a proposal that is still open.`.

I think the solution would be to better handle the error messages: https://github.com/TaggrNetwork/Taggr/pull/6/files

mechaquan:
So you mean like more specific error responses?  Do we know what other possibilities are?  Low balance, open proposal, etc... what else?

`Transfer failed with error: ` + response ... is there a list of what these responses are?  Sorry I need GitHub for dummies...

leo-ship-it:
So the error you get is of type:
```
type TransferError = variant {
  GenericError : GenericError;
  BadFee : BadFee;
  CreatedInFuture : CreatedInFuture;
  TooOld;
  InsufficientFunds : InsufficientFunds;
};
```

The response you get is of type one of those. 
Generic error can be one of:
- minting invariant violation
- only minting transactions are allowed without a fee
- transfers locked: a vote on a pending proposal detected
- memo longer than 32 bytes

xqxpx (approved):

xqxpx:
Thanks Leo!

PR #5, opened 2024-02-18
feat: block based allocation

xqxpx:
## Problem

The current memory allocator (only used for storing posts in the stable memory so far) is very simple and works on the best effort basis when trying to allocate new memory. It keeps a set of free segments and on every new allocation it simply finds the next segment that is large enough. This was very easy to build and worked great so far. The only problem is that this strategy fragments the stable memory and leaves a lot of small holes which cannot be used for new posts and makes the search of the next free segment longer and longer. 

## Proposal

Switch to a block based allocation and allocate memory in small blocks. This had the disadvantage that we are slightly more wasteful with resources, however, if we chose the size of the block large enough to fit the smallest storable unit, the allocator will automatically close all gaps so that the fragmentation will never be a problem.

By doing some measurements I found out that _most posts_ are slightly below 300 bytes. By choosing `300` bytes as the block size, many newly archived posts will eventually close all gaps that will appear when older posts are moved back to heap for editing.

## PR

In this PR we:

- restore the state and set the block size to 1 byte (it's effectively the current "block" size)
- do a no-op editing of all posts and effectively move them from the stable memory to the heap (~111Mb)
- configure the new allocator to use 300 bytes as the block size.

After this change is released, Taggr will daily move 50k posts back to the stable memory using the new allocator setting.

This change was tested with the mainnet data.

## Testing

Testing on local replica with the mainnet data shows very promising results. What I did:

1. Upgraded to this branch. (all posts moved to heap)
2. Archived a big chunk of posts (From JS console: `api.call("archive", 40000)`).
3. Started randomly modify any posts in the range 0-40k.
4. Make sure that free segments started to appear (From JS console: `api.query("stats").then(x=>console.log(x.meta))`)
5. Archive some more posts (From JS console: `api.call("archive", 100)`).
6. Look at the number of segments and verify that we're back to `1` segment.


src/backend/env/memory.rs
```diff
@@ -179,7 +197,12 @@ impl Default for Allocator {
 }
 
 impl Allocator {
-    fn alloc(&mut self, n: u64) -> Result<u64, String> {
+    fn get_allocation_length(&self, n: u64) -> u64 {
+        (n + self.block_size_bytes - 1) / self.block_size_bytes * self.block_size_bytes
```
leo-ship-it, line 201:
```suggestion
        (n + self.block_size_bytes - 1) / self.block_size_bytes.pow(2)
```
Where does this formula come from? Also, how do we know that `self.block_size_bytes` is never going to be 0?

taggrx: The suggestion would break the tests because if `/` is an integer division, then `13/4 * 4 = 3 * 4` while `13/16 = 0`.

taggrx: The formula is from me. I just need an upscaling integer division.

PR #6, opened 2024-02-18
Improve transfer error

leo-ship-it:


src/frontend/src/wallet.tsx
```diff
@@ -496,9 +496,7 @@ export const Wallet = () => {
                             transaction_fee,
                         );
                         if (typeof response == "string")
-                            alert(
-                                `Transfer failed. One reason might be that you voted on a proposal that is still open.`,
-                            );
+                            alert(`Transfer failed with error: ` + response);
```
taggrx, line 499: No this will break the error message. Currently, the Taggr frontend does not parse ICRC1 errors, so there is no way to know why the transfer has failed. The `response` contains some generic error message.

So this change should not be merged.

PR #8, opened 2024-02-19
Release

xqxpx:
- Fixes realm revenue number in reports
- Refactors tipping code to be synchronous
- Fixes search by using indexes
- Fixes timestamp values in metrics
- Fixes overcharging of reaction fees on reports
- Fixes scrolling position on posts


src/backend/env/mod.rs
```diff
@@ -1625,16 +1623,15 @@ impl State {
...
                 .filter_map(|user_id| self.users.get(&user_id))
                 .filter(|user| user.active_within_weeks(now, 1))
                 .map(|user| (user.id, user.name.clone()))
                 .collect::<Vec<_>>();
-            let controller_revenue = revenue * CONFIG.realm_revenue_percentage as u64
-                / 100
-                / controllers.len().max(1) as u64;
+            let realm_revenue = revenue * CONFIG.realm_revenue_percentage as u64 / 100;
+            total_revenue += realm_revenue;
+            let controller_revenue = realm_revenue / controllers.len().max(1) as u64;
```
leo-ship-it, line 1634: Does this mean that previously total_revenue revenue was wrong?

taggrx: Yes! But this is only used for reports (displaying), not for distribution. @radu-daniel found the issue.

PR #9, opened 2024-02-19
fix: archive old posts

xqxpx:


src/backend/env/mod.rs
```diff
@@ -1393,8 +1393,8 @@ impl State {
         export_token_supply(token::icrc1_total_supply());
     }
 
-    fn archive_cold_data(&mut self) -> Result<(), String> {
-        let max_posts_in_heap = 50_000;
+    pub fn archive_cold_data(&mut self) -> Result<(), String> {
+        let max_posts_in_heap = 10_000;
```
ulan, line 1397: With this decrease, archive_cold_posts will do more work right? How do we know that it will not run out of instructions? Or do you want to try it?

ulan: Do know which of the chores might be causing out-of-instructions?

taggrx: `archive_cold_data` takes this constant as a parameter and then archives exactly that amount of posts, on purpose to not run out of instructions. Since we are archive above 10k for a month, I think this is enough, unless there is a bug of unrestricted post length and somebody actually did it (I'll check this later once the problem is fixed).

Regarding your second question, the hourly one is _definitely_ crashing since the last backups are from 5h ago.

What I cannot explain is why the `daily_chores` routine is not cleaning up. The way the code is written, Taggr should try execute it before the hourly one and then it would clean up. This was one of the reasons for this ordering. My suspicion is that also `archive_cold_data` became too expensive _at the same time_ so that in reality it is actually this method which is crashing.

taggrx: Actually, let me parameterize this method, so that it can be called with any number.

taggrx: Btw, I suspect the issue is the memory fragmentation which makes the posts insertion into the stable memory more and more expensive.

ulan: > archive_cold_data takes this constant as a parameter and then archives exactly that amount of posts, on purpose to not run out of instructions.

The reason I asked is because the parameter specifies how many posts to keep in memory, not how many posts to archive. So decreasing the parameter, actually increases the number of posts to archive.

Adding an additional parameter for how many posts to archive sgtm!

taggrx: > The reason I asked is because the parameter specifies how many posts to keep in memory, not how many posts to archive. 

True, this parameter is simply used for both things - to limit how much we archive at once and how much we keep in the heap. There is no good explanation, I guess I didn't want to think about the value of a second parameter as this number should be anyway small enough to handle without issues.

I just tested the change locally, it works as expected.

ulan: Oh, I missed the part where it takes the min of this parameter. Sg!

src/backend/updates.rs
```diff
@@ -1322,3 +673,19 @@ fn check_candid_interface_compatibility() {
...
     );
 }
+
+#[export_name = "canister_update archive"]
+fn archive() {
+    mutate(|state| {
+        if !state.principal_to_user(caller()).unwrap().stalwart {
+            return;
+        }
+        let n: usize = parse(&arg_data_raw());
```
ulan, line 683: Even if a stalwart is malicious, then cannot do any harm here, right? Too small n and too large n both work properly.

ulan:
lgtm!

PR #14, opened 2024-02-22
Data migration, Part I: move posts to heap and remove archiving

xqxpx:
This PR is the first part of Taggr's migration to DFINITY stable-structure library.

This change makes Taggr move all posts to the heap from the stable memory and disables the archiving.

## Reason

The current stable memory allocator was implemented by X approximately a year ago, when no comparable solutions existed that could match all requirements of Taggr. This solution worked well until today, but with the growing data volume of Taggr, we started experiencing problems, that require further development efforts. 

In the mean time, DFINITY came up with a very mature and promising library, that can address the current trade offs of Taggr and goes even beyond that. This migration will give us peace of mind for the foreseeable future.


src/backend/updates.rs
```diff
@@ -70,7 +70,19 @@ fn post_upgrade() {
     });
 }
 
-async fn post_upgrade_fixtures() {}
+async fn post_upgrade_fixtures() {
+    mutate(|state| {
```
ulan, line 74: Shouldn't this be part of `post_upgrade` such that if it fails, then the entire upgrade fails?

taggrx: From our previous discussions about instruction limits I understood that if it's not an actual canister call, the execution happens within the same context. Did I misunderstand this? Because that's indeed what I assume here: that `post_upgrade_fixtures` is "inlined" into `post_upgrade`.

ulan: It is called through a timer:
```
set_timer(Duration::from_millis(0), move || {
        spawn(post_upgrade_fixtures())
    });
```

So it will run when the timer is scheduled to run.

taggrx: But it is subject to the same instruction budged as the parent call? 

PS: Thanks! Ok, then I'll definitely properly inline!

ulan: The instruction budget is different as well: upgrade => 200B (for pre/post upgrade combined), timer => 20B.

PR #15, opened 2024-02-22
Data migration, Part II: init stable structures and move posts to it

xqxpx:
In this change we introduce a second step of data migration to the stable-structures library:

1. In post-upgrade, after we loaded the heap, we initialize all data structures (`POSTS` and `HEAP`). This is the reason why we need an `Option` right now (it can be removed later).
2. After the upgrade we put the state into a read only mode which only affects the data we're moving: the posts. During the migration, users will not be able to create or modify posts.
3. Then, Taggr calls itself multiple time to move posts in chunks of 20k posts into the new data structure.
4. If no posts are left in the heap, Taggr switches off the read only mode and the migration counts as successful. If something goes wrong, Taggr stops the migration and leaves the state in read only mode. In that case, we will need analyze the situation and do a recovery upgrade.

### DIFF
--- a/src/backend/env/post.rs
+++ b/src/backend/env/post.rs
@@ -1,30 +1,16 @@
-use std::cmp::{Ordering, PartialOrd};
-
 use super::reports::ReportState;
 use super::*;
 use super::{storage::Storage, user::UserId};
 use crate::mutate;
 use crate::reports::Report;
+use ic_stable_structures::{StableBTreeMap, Storable};
 use serde::{Deserialize, Serialize};
+use std::borrow::Cow;
+use std::cell::RefCell;
+use std::cmp::{Ordering, PartialOrd};
 
-static mut CACHE: Option<BTreeMap<PostId, Box<Post>>> = None;
-
-// This is a static cache that is only populated during one queries's life cycle. The reason why it's
-// needed is that `Post::get` always returns a reference to a post. But when we load an archived
-// post, we read raw bytes from the stable memory and deserialize them into a value. To return a
-// reference to that value, we need to anchor it somewhere on the heap. This is where we need our
-// cache. But we also _cannot_ store posts inside the cache directly, because any mutation of the
-// cache within a query life cycle will restructure the hash map and hence break all references.
-// To work around this issue, we go through one level of indirection and box all posts (put them on
-// to the heap) and then only add the pointers to the boxed value into the cache. This way,
-// when we get a reference to a post and dereference twice, we get a stable reference.
-fn cache<'a>() -> &'a mut BTreeMap<PostId, Box<Post>> {
-    unsafe {
-        if CACHE.is_none() {
-            CACHE = Some(Default::default())
-        }
-        CACHE.as_mut().expect("no cache instantiated")
-    }
+thread_local! {
+    pub static POSTS: RefCell<Option<StableBTreeMap<PostId, Post, crate::Memory>>> = Default::default();
 }
 
 pub type PostId = u64;
@@ -77,6 +63,19 @@ pub struct Post {
 
     #[serde(skip)]
     pub archived: bool,
+}
+
+impl Storable for Post {
+    const BOUND: ic_stable_structures::storable::Bound =
+        ic_stable_structures::storable::Bound::Unbounded;
+
+    fn to_bytes(&self) -> Cow<[u8]> {
+        Cow::Owned(serde_cbor::to_vec(self).expect("post serialization failed"))
+    }
+
+    fn from_bytes(bytes: Cow<[u8]>) -> Self {
+        serde_cbor::from_slice(&bytes).expect("post deserialization failed")
+    }
 }
 
 impl PartialEq for Post {
@@ -219,11 +218,9 @@ impl Post {
     }
 
     pub async fn save_blobs(post_id: PostId, blobs: Vec<(String, Blob)>) -> Result<(), String> {
-        let existing_blobs = read(|state| {
-            Post::get(state, &post_id)
-                .map(|post| post.files.keys().cloned().collect::<BTreeSet<_>>())
-        })
-        .unwrap_or_default();
+        let existing_blobs = Post::get(&post_id)
+            .map(|post| post.files.keys().cloned().collect::<BTreeSet<_>>())
+            .unwrap_or_default();
 
         for (id, blob) in blobs
             .into_iter()
@@ -336,7 +333,7 @@ impl Post {
     /// Checks if the poll has ended. If not, returns `Ok(false)`. If the poll ended,
     /// returns `Ok(true)` and assings the result weighted by the token voting power.
     pub fn conclude_poll(state: &mut State, post_id: &PostId, now: u64) -> Result<bool, String> {
-        let user_balances = Post::get(state, post_id)
+        let user_balances = Post::get(post_id)
             .and_then(|post| {
                 if let Some(Extension::Poll(poll)) = post.extension.as_ref() {
                     let user_ids = poll.votes.values().flatten().cloned();
@@ -382,7 +379,7 @@ impl Post {
     ) -> Result<(), String> {
         mutate(|state| {
             let user = state.principal_to_user(principal).ok_or("no user found")?;
-            let mut post = Post::get(state, &id).ok_or("no post found")?.clone();
+            let mut post = Post::get(&id).ok_or("no post found")?.clone();
             if post.user != user.id {
                 return Err("unauthorized".to_string());
             }
@@ -416,10 +413,7 @@ impl Post {
 
             let current_realm = post.realm.clone();
 
-            // After we validated the new edited copy of the post, charged the user, we should remove the
-            // old post, and insert the edited one.
-            Post::take(state, &id);
-            Post::save(state, post);
+            Post::save(post);
 
             if current_realm != picked_realm {
                 change_realm(state, id, picked_realm)
@@ -443,6 +437,9 @@ impl Post {
         picked_realm: Option<String>,
         extension: Option<Extension>,
     ) -> Result<PostId, String> {
+        if state.read_only {
+            return Err("read-only mode".into());
+        }
         let user = match state.principal_to_user(principal) {
             Some(user) => user,
             // look for an authorized controller
@@ -463,7 +460,7 @@ impl Post {
             return Err("bots can't create comments".into());
         }
 
-        let realm = match parent.and_then(|id| Post::get(state, &id)) {
+        let realm = match parent.and_then(|id| Post::get(&id)) {
             Some(parent) => parent.realm.clone(),
             None => match picked_realm {
                 Some(value) if value.to_lowercase() == CONFIG.name.to_lowercase() => None,
@@ -487,7 +484,7 @@ impl Post {
             }
         } else if let Some(discussion_owner) = parent.and_then(|post_id| {
             state.thread(post_id).next().and_then(|post_id| {
-                Post::get(state, &post_id).and_then(|post| state.users.get(&post.user))
+                Post::get(&post_id).and_then(|post| state.users.get(&post.user))
             })
         }) {
             if !discussion_owner.accepts(user.id, &user.get_filter()) {
@@ -516,7 +513,7 @@ impl Post {
         let future_id = state.next_post_id;
         let is_comment = parent.is_some();
         let excess_factor = user
-            .posts(state, 0, is_comment)
+            .posts(0, is_comment)
             .take_while(|post| post.timestamp() + if is_comment { HOUR } else { DAY } > timestamp)
             .count()
             .saturating_sub(if is_comment {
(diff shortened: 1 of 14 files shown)


src/backend/env/post.rs
```diff
@@ -621,52 +618,33 @@ impl Post {
...
-                    let cache = cache();
-                    post.archived = true;
-                    cache.insert(*post_id, Box::new(post));
-                    cache.get(post_id)
-                })
-            });
-            boxed.map(|ptr| &**ptr)
-        })
+    pub fn get(post_id: &PostId) -> Option<Post> {
+        POSTS.with(|p| p.borrow().as_ref().unwrap().get(post_id))
```
ulan, line 626: Shouldn't we also try getting the posts from `state.posts`? Otherwise, the system will not show any posts until migration is done, right?

src/backend/updates.rs
```diff
@@ -71,19 +78,93 @@ fn post_upgrade() {
...
-                post.archived = false;
-                Ok(())
-            })
-            .unwrap();
+        let mut r = 0;
+        for _ in 0..20000 {
+            let Some((_, post)) = state.posts.pop_first() else {
+                break;
+            };
+            Post::save(post);
```
ulan, line 133: Would it make set to `Post::get()` and compare with the old value to make sure that the post was saved correctly? If the comparison fails, then move the post to a new `state.failed_posts` map?

src/backend/updates.rs
```diff
@@ -71,19 +78,93 @@ fn post_upgrade() {
...
+fn migrate_posts() {
     mutate(|state| {
-        let num_posts = Post::count(state);
-        for id in 0..state.next_post_id {
-            Post::mutate(state, &id, |post| {
-                post.archived = false;
-                Ok(())
-            })
-            .unwrap();
+        let mut r = 0;
```
ulan, line 128: How many instructions does this loop take at most locally? Are we far away from 20B. In terms of performance it might be better to keep this number below 2B (e.g. 1.5B).

src/backend/updates.rs
```diff
@@ -71,19 +78,93 @@ fn post_upgrade() {
...
+                    ))
+                })
+            });
+        let new_num_posts = read(|state| state.posts.len());
+        if new_num_posts == 0 {
+            // We're done with migration
+            break;
+        }
+        if num_posts == new_num_posts && num_posts > 0 {
+            // Something went wrong, so we stop and let the state locked.
```
ulan, line 105: Wdyt about moving the failed posts into a new `state.failed_posts` map and then continue migration with the remaining posts? Once the migration is done, we could move `state.failed_posts` to `state.posts`, clear `read_only` flag and attempt a new upgrade via DAO even if some posts have failed (assuming the number of failed posts is low enough)?

src/backend/updates.rs
```diff
@@ -71,19 +78,93 @@ fn post_upgrade() {
...
+            state.read_only = false;
+            state
+                .logger
+                .debug(format!("Succefully migrated {} posts", total_posts))
+        })
+    }
+}
+
+#[update]
+fn migrate_posts() {
```
ulan, line 126: Would it make sense to make 20000 into a parameter for flexibility?

src/backend/updates.rs
```diff
@@ -71,19 +78,93 @@ fn post_upgrade() {
...
+
+    if total_posts as u64 == Post::count() {
+        mutate(|state| {
+            state.read_only = false;
+            state
+                .logger
+                .debug(format!("Succefully migrated {} posts", total_posts))
+        })
+    }
+}
```
ulan, line 123: Let's also introduce an endpoint that copies the body of the loop above. I.e. represents one step of migration. This would allow us to complete the migration manually even if the loop above stops for some reason.

src/backend/updates.rs
```diff
@@ -71,19 +78,93 @@ fn post_upgrade() {
 }
 
 async fn post_upgrade_fixtures() {
+    mutate(|state| state.read_only = true);
```
ulan, line 81: I think we should set the `read_only` flag in the `post_upgrade` to make sure that we start in read-only mode after the upgrade without having any window of time where the flag is false.

src/backend/updates.rs
```diff
@@ -71,19 +78,93 @@ fn post_upgrade() {
...
+    let total_posts = read(|state| state.posts.len());
+
+    // Init all stable structs becasue we do not need the old contents of the stable memory anymore
+    init_stable_structs();
```
ulan, line 85: Let's move it to post-upgrade to make sure that stable structs are initialized after upgrade even if this function traps.

ulan:
> In that case, we will need analyze the situation and do a recovery upgrade.

Does the recovery upgrade work even if the read-only mode is one?

src/backend/updates.rs
```diff
@@ -41,7 +42,13 @@ fn init() {
 
 #[pre_upgrade]
 fn pre_upgrade() {
-    mutate(env::memory::heap_to_stable)
+    HEAP.with(|cell| {
```
ulan, line 45: How many instructions does this take with the production state?

ulan:
Let's collect some test scenarios we should try locally?

1. Upgrade twice with the production data (to make sure that pre-upgrade works).
2. Introduce artificial failure for some posts during migration and make sure that upgrading is still possible.
3. Introduce artificial failure in the migration loop and make sure that we can complete migration manually by calling the endpoints.
4. Measure how the number of instructions and stable memory size grows for serialization of the state to stable memory with the number of transactions.
5. Measure how the stable memory grows if a post is modified many times.
6. Measure how the stable memory grows with many back to back upgrades (to make sure the serialization of the state can reuse stable memory and there is no leak).

Anything else?

src/backend/env/post.rs
taggrx: That's correct. But the migration is expected to last only a couple of minutes. Is it worth adding complexity to bridge this downtime?

src/backend/updates.rs
taggrx: Definitely!

src/backend/updates.rs
taggrx: I didn't measure yet, but serializing 20k posts in one single call that does not do anything is not an issue currently. I hope stable-structs lib does not make it much worse :). But for local tests, I'll add a print line with the number of instructions - thanks for the idea!

src/backend/updates.rs
taggrx: Yes, this is a great idea, will add!

src/backend/updates.rs
taggrx: How would this parameter be controlled? This entire call is one-off code, that gets removed in the next PR. Maybe I misunderstand your intention?

src/backend/updates.rs
taggrx: It is already a callable endpoint, right?

src/backend/updates.rs
taggrx: I agree, but this goes to back my question on the first PR. I assumed that this function is inlined. I'll wait for your response on the other PR and if it's not inline I simply remove `async` and `spawn` when calling the fixture functions.

xqxpx:
> > In that case, we will need analyze the situation and do a recovery upgrade.
> 
> Does the recovery upgrade work even if the read-only mode is one?

Yes! Because the recovery upgrade (on purpose) does not depend on the ability to create posts.

src/backend/updates.rs
taggrx: Will check. But since this will triggered only after a successful migration, we would have no posts in the heap, and the heap size will be around 23MB.

When the migration fails for some reason and all posts stayed in the heap, I can't see many options without deleting posts, can you?

src/backend/env/post.rs
ulan: It is a couple of lines of code that would keep the webpage functioning in read-only mode: if the post is in state.posts then return it otherwise use the current code?

src/backend/updates.rs
ulan: I was thinking about
```
#[update]
fn migration_step(num_posts_to_migrate) {
  // The body of the loop that call internally `migrate_posts(num_posts_to_migrate)`
  // but also has the logic of moving failed posts to `state.failed_posts` and logic to
  // complete migration if there are no posts remaining (copying `state.failed_posts` back to `state.posts`)
}
```

src/backend/updates.rs
ulan: I am thinking if for some reason 20000 runs out of instructions, we could call it with smaller number (see my comment about `migration_step` above).

xqxpx:
> 1. Upgrade twice with the production data (to make sure that pre-upgrade works).

Not really yet. This PR doesn't touch restoring of the heap. This is done in the third PR. So what you can do right now (on mainnet data!) Is sync backup to your local replica and then deploy 3 PRs one after the other. This proves that the happy path works.

> 2. Introduce artificial failure for some posts during migration and make sure that upgrading is still possible.
> 3. Introduce artificial failure in the migration loop and make sure that we can complete migration manually by calling the endpoints.

Yes! And the most important test is to make sure that the recovery upgrade works in "read_only" mode. 

> 4. Measure how the number of instructions and stable memory size grows for serialization of the state to stable memory with the number of transactions.

Could you please explain this one? Which "transactions"?

> 5. Measure how the stable memory grows if a post is modified many times.
> 6. Measure how the stable memory grows with many back to back upgrades (to make sure the serialization of the state can reuse stable memory and there is no leak).

Haha, ok now I am a bit worried to see these suggestions... 😅 If we use the official API, do you something can go wrong with that? The point is that local measurements will not prove the absence of future issues in the library? And I would not expect, e.g. that `Cell` grows the stable memory every time I write into it, because how would it be different from a `Log`?

src/backend/updates.rs
ulan: It runs in timer, not in post_upgrade. Note that this function cannot be inlined in post_upgrade because this function makes calls (post_upgrade is not allowed to make calls).

I was suggesting to move the sync code into post_upgrade. We could also try to do as many migration steps as possible (synchronously) inside post_upgrade. And assert that `state.failed_posts` is empty. That has a benefit that if something goes wrong in those steps, the entire upgrade will be reverted by the system, so the canister remains in a clean state.

ulan:
> Could you please explain this one? Which "transactions"?

I meant the ledger transactions to make sure that the growth of `HEAP` is expected.

> Haha, ok now I am a bit worried to see these suggestions... 😅 If we use the official API, do you something can go wrong with that?

This is intended more as a sanity check that we are using the library correctly. (I used the library only in toy projects, so I am not an expert there).

ulan:
> Not really yet. This PR doesn't touch restoring of the heap. This is done in the third PR.

I understood that the plan is to release PR I and PR II as a single release, right? And then follow up with PR III in another release. For each release, ideally the upgrades are idenmpotent (i.e. it doesn't matter how many times upgrades are done).

xqxpx:
> I understood that the plan is to release PR I and PR II as a single release, right? 

No, actually I thought about doing two releases, to keep PRs and mental complexity small. But if you think the complexity would be still managable, how about:

1. Combining PRs I & II.
2. Making Taggr work in read-only mode (your suggestion above).
3. Simplify the migration to a fully manual (controllable by any stalwart) state lock/unlock & pagination.

Then we upgrade Taggr and gather in a _public_ (!) OC chat (for ultimate transparency) and then in a controlled way:

1. Lock the state (actually only post modifications/creations).
2. Make enough _parameterized_ calls to move posts to heap (Taggr will log every step).
3. Initialize stable structs.
4. Make enough parameterized calls to move all posts back to stable mem.
5. Unlock the state.

There needs to be a simple state machine logic to not allow interleaving actions. Any of the steps will be callable by any stalwart from the JS console.

This should take not more than 5 minutes in total.

WDYT?

ulan:
If we would do PR I as a standalone release, then we might not be able to upgrade from it because all posts will be in memory and pre-upgrade would run out of instructions, right? Maybe I am missing something.

> 1. Combining PRs I & II.
> 2. Making Taggr work in read-only mode (your suggestion above).

These steps sgtm!

I think it would be safe to do as much of the migration in upgrade as possible, because if anything traps there, then the entire upgrade will fail so the canister state will remain consistent.

Alternative plan:
1. In post-upgrade: move posts to heap, lock the state, initialize stable structs, migrate as many posts synchronously as the instruction limit permits (trapping in case of a failure). 
2. Make enough parameterized calls to migrate the remaining posts.
3. Unblock the state.

xqxpx:
> If we would do PR I as a standalone release, then we might not be able to upgrade from it because all posts will be in memory and pre-upgrade would run out of instructions, right? Maybe I am missing something.

On a local replica, at least with current post numbers it still works without issues. But the longer we wait, the less difficult it might become indeed.

I like the alternative plan, but:

>migrate as many posts synchronously as the instruction limit permits

If the total migration is not possible with leaving a significant margin and additional calls are necessary, I wouldn't try to split hairs here and just migrate posts in subsequent calls.

Ok, let me combine the PRs first and apply all things we discussed including measurements.

PR #16, opened 2024-02-22
Data migration, Part II: finalization of the migration

xqxpx:
In this PR we restore the heap from the new place and remove all code that is not needed anymore.

This change does not need to be deployed quickly.


beiji-zz:
Can anyone help me with this problem?
https://forum.dfinity.org/t/does-anyone-know-how-to-get-taggrs-transaction-history/27895

PR #20, opened 2024-02-23
Add PocketIC testing

ulan:
This adds a new `tests` directory with PocketIC integration tests.

Summary of changes:

- Add a script to download `pocket-ic` binary (tested only on linux).
- Add a `make pocket_ic` rule.
- Adjust the `make test` rule to first build the Wasm binaries and call `cargo test`.
- Add tests for the backend and bucket canisters.

### DIFF
--- a/tests/src/backend_tests.rs
+++ b/tests/src/backend_tests.rs
@@ -1,24 +1,40 @@
+use std::path::Path;
+
 use candid::{decode_one, encode_args, encode_one, Principal};
 use ic_cdk::api::management_canister::main::CanisterId;
-use pocket_ic::{PocketIc, WasmResult};
+use pocket_ic::{ErrorCode, PocketIc, WasmResult};
 
-use crate::{controller, get_wasm, setup};
+use crate::{controller, get_wasm, setup, setup_from_snapshot};
 
-fn create_test_user(pic: &PocketIc, backend: CanisterId, caller: Principal, name: &str) -> u64 {
-    let result = pic
-        .update_call(
-            backend,
-            caller,
-            "create_test_user",
-            encode_one(name).unwrap(),
-        )
-        .unwrap();
+fn create_test_user(
+    pic: &PocketIc,
+    backend: CanisterId,
+    caller: Principal,
+    name: &str,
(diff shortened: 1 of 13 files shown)


ulan:
@taggrx: ptal

tests/src/backend_tests.rs
```diff
@@ -0,0 +1,131 @@
...
+    let _user_id = create_test_user(&pic, backend, controller(), "test");
+    let post_id = add_post(&pic, backend, controller(), "lorem ipsum", vec![], None).unwrap();
+    let post = &posts(&pic, backend, controller(), vec![post_id])[0];
+
+    assert_eq!(post.get("id").unwrap().as_u64().unwrap(), post_id);
+    assert_eq!(post.get("body").unwrap().as_str().unwrap(), "lorem ipsum");
+}
+
+#[test]
+fn test_upgrades() {
```
taggrx, line 149: I would definitely keep this test as it serves as a great example for me when I'll add more of these tests, but I wonder if you knew that our e2e tests do the same but via UI and even doing a second upgrade using the recovery mechanism?

xqxpx (approved):
Thanks so much Ulan, this is was indeed a missing piece so far: there is functionality which cannot be really tested via e2e tests like distributions and cannot be tested as a unit test. 

Do you agree that we should use these tests especially for this kind of functionalities?

ulan:
> Thanks so much Ulan, this is was indeed a missing piece so far: there is functionality which cannot be really tested via e2e tests like distributions and cannot be tested as a unit test.
> 
> Do you agree that we should use these tests especially for this kind of functionalities?

I am hoping this sits between unit tests and e2e tests and allows us to test all public endpoints. I know that some of the tests will be duplicated, but this allows us to exercise conditions that UI wouldn't allow (since UI also has its own checks, but the attacker doesn't have to use the UI).

Another benefit of this framework is that it will be way faster that e2e tests.

xqxpx:
Btw, I finally fixed the Docker issue, feel free to merge from `main`.

xqxpx (approved):

PR #21, opened 2024-02-24
Part I: migrate archived posts to stable-structs

xqxpx:
This PR introduces a new stable memory allocator based on stable structs and moves all posts from the old allocator into the heap first.

If the upgrade procedure succeeds, the canister will initialize the new allocator. Starting from this point, the cold-archiving will happen manually moving 30k posts (6B instructions) to the new storage hourly. If necessary, the `archive` method will be still available for manual intervention.

### DIFF
--- a/src/backend/env/mod.rs
+++ b/src/backend/env/mod.rs
@@ -10,15 +10,18 @@ use crate::env::user::CreditsDelta;
 use crate::proposals::Proposal;
 use crate::token::{Account, Token, Transaction};
 use crate::{assets, mutate, read};
+use crate::{id, time};
 use candid::Principal;
 use config::{CONFIG, ICP_CYCLES_PER_XDR};
 use ic_cdk::api::stable::stable64_size;
 use ic_cdk::api::{self, canister_balance};
 use ic_ledger_types::{AccountIdentifier, DEFAULT_SUBACCOUNT, MAINNET_LEDGER_CANISTER_ID};
+use ic_stable_structures::Storable;
 use invoices::Invoices;
 use serde::{Deserialize, Serialize};
 use serde_bytes::ByteBuf;
 use sha2::{Digest, Sha256};
+use std::borrow::Cow;
 use std::cmp::Reverse;
 use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet, VecDeque};
 use user::{User, UserId};
@@ -194,7 +197,6 @@ pub struct State {
 
     migrations: BTreeSet<UserId>,
 
-    #[serde(default)]
     pub posts_with_tags: Vec<PostId>,
 }
 
@@ -246,6 +248,19 @@ impl Logger {
 pub enum Destination {
     Rewards,
     Credits,
+}
+
+impl Storable for State {
+    const BOUND: ic_stable_structures::storable::Bound =
+        ic_stable_structures::storable::Bound::Unbounded;
+
+    fn to_bytes(&self) -> Cow<[u8]> {
+        Cow::Owned(serde_cbor::to_vec(self).expect("state serialization failed"))
+    }
+
+    fn from_bytes(bytes: Cow<[u8]>) -> Self {
+        serde_cbor::from_slice(&bytes).expect("state deserialization failed")
+    }
 }
 
 impl State {
@@ -1367,16 +1382,6 @@ impl State {
                 state.logger.info("An emergency release is pending! 🚨");
             }
 
-            if let Err(err) = state.archive_cold_data() {
-                state
-                    .logger
-                    .error(format!("couldn't archive cold data: {:?}", err));
-            } else {
-                // If we archived posts without errors, we should create a heap backup, so that we
-                // always have a valid heap snapshot at the end of the stable memory.
-                memory::heap_to_stable(state);
-            }
-
             state.recompute_stalwarts(now);
 
             for user in state.users.values_mut() {
@@ -1392,7 +1397,7 @@ impl State {
     }
 
     fn archive_cold_data(&mut self) -> Result<(), String> {
-        let max_posts_in_heap = 10_000;
+        let max_posts_in_heap = 30_000;
         archive_cold_posts(self, max_posts_in_heap)
     }
 
@@ -1514,9 +1519,11 @@ impl State {
 
     pub async fn hourly_chores(now: u64) {
         mutate(|state| {
-            // Automatically dump the heap to the stable memory. This should be the first
-            // opearation to avoid blocking of the backup by a panic in other parts of the routine.
-            memory::heap_to_stable(state);
+            if let Err(err) = state.archive_cold_data() {
+                state
+                    .logger
+                    .error(format!("couldn't archive cold data: {:?}", err));
+            }
 
             state.conclude_polls(now)
         });
@@ -2746,20 +2753,6 @@ fn covered_by_feeds(
         }
     }
     None
-}
-
-pub fn id() -> Principal {
-    #[cfg(test)]
-    return Principal::anonymous();
-    #[cfg(not(test))]
-    ic_cdk::id()
-}
-
-pub fn time() -> u64 {
-    #[cfg(test)]
-    return 0;
-    #[cfg(not(test))]
-    api::time()
 }
 
 pub fn parse_amount(amount: &str, token_decimals: u8) -> Result<u64, String> {
--- a/src/backend/env/post.rs
+++ b/src/backend/env/post.rs
@@ -654,12 +654,8 @@ impl Post {
     pub fn get<'a>(state: &'a State, post_id: &PostId) -> Option<&'a Post> {
         state.posts.get(post_id).or_else(|| {
             let boxed = cache().get(post_id).or_else(|| {
-                // TODO: delete state.memory access after migration
-                state
-                    .memory
-                    .posts
-                    .get(post_id)
-                    .or_else(|| POSTS.with(|p| p.borrow().as_ref().unwrap().get(post_id)))
+                POSTS
+                    .with(|p| p.borrow().as_ref().unwrap().get(post_id))
                     .and_then(|mut post: Post| {
                         let cache = cache();
                         post.archived = true;
@@ -750,9 +746,6 @@ pub fn archive_cold_posts(state: &mut State, max_posts_in_heap: usize) -> Result
                 .posts
                 .remove(&post_id)
                 .ok_or(format!("no post found for id={post_id}"))?;
-            #[cfg(test)]
-            state.memory.posts.insert(post_id, post.clone()).unwrap();
-            #[cfg(not(test))]
             POSTS.with(|p| p.borrow_mut().as_mut().unwrap().insert(post.id, post));
             Ok::<(), String>(())
         })
@@ -939,6 +932,7 @@ mod tests {
         STATE,
     };
 
+    #[ignore]
     #[test]
     fn test_post_archiving() {
         static mut MEM_END: u64 = 16;
--- a/src/backend/env/post.rs
+++ b/src/backend/env/post.rs
@@ -1,11 +1,17 @@
-use std::cmp::{Ordering, PartialOrd};
-
 use super::reports::ReportState;
 use super::*;
 use super::{storage::Storage, user::UserId};
-use crate::mutate;
 use crate::reports::Report;
+use crate::{mutate, performance_counter};
+use ic_stable_structures::{StableBTreeMap, Storable};
 use serde::{Deserialize, Serialize};
+use std::borrow::Cow;
+use std::cell::RefCell;
+use std::cmp::{Ordering, PartialOrd};
+
+thread_local! {
+    pub static POSTS: RefCell<Option<StableBTreeMap<PostId, Post, crate::Memory>>> = Default::default();
+}
 
 static mut CACHE: Option<BTreeMap<PostId, Box<Post>>> = None;
 
@@ -88,6 +94,19 @@ impl PartialEq for Post {
 impl PartialOrd for Post {
     fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
         Some(self.id.cmp(&other.id))
+    }
+}
+
+impl Storable for Post {
+    const BOUND: ic_stable_structures::storable::Bound =
+        ic_stable_structures::storable::Bound::Unbounded;
+
+    fn to_bytes(&self) -> Cow<[u8]> {
+        Cow::Owned(serde_cbor::to_vec(self).expect("post serialization failed"))
+    }
+
+    fn from_bytes(bytes: Cow<[u8]>) -> Self {
+        serde_cbor::from_slice(&bytes).expect("post deserialization failed")
     }
 }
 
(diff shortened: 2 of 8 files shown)


src/backend/env/post.rs
```diff
@@ -710,13 +750,20 @@ pub fn archive_cold_posts(state: &mut State, max_posts_in_heap: usize) -> Result
                 .posts
                 .remove(&post_id)
                 .ok_or(format!("no post found for id={post_id}"))?;
-            state.memory.posts.insert(post_id, post)
+            #[cfg(test)]
```
ulan: I think we can avoid it by initializing `POST` in test. Stable structures in test mode should use an in-memory storage automatically.

src/backend/updates.rs
```diff
@@ -64,12 +71,60 @@ fn post_upgrade() {
...
 
+    migrate_posts();
+
     // post upgrade logic goes here
     set_timer(Duration::from_millis(0), move || {
         spawn(post_upgrade_fixtures());
     });
 }
 
+fn migrate_posts() {
```
ulan, line 83: Just to double check: we still have the old end-point that is callable by stalwarts that can trigger this function manually, right?

taggrx: Yes! This is the `archive` endpoint I mention in the description!

xqxpx:
Btw, I see some upgrade tests failing, but this is unavoidable as we change the heap serialization as well, I think this tests can be ignored (or disabled).

src/backend/env/post.rs
ulan: Or is it done to avoid modifying the tests? I am confused.

src/backend/env/post.rs
```diff
@@ -622,20 +641,31 @@ impl Post {
...
-                    let cache = cache();
-                    post.archived = true;
-                    cache.insert(*post_id, Box::new(post));
-                    cache.get(post_id)
-                })
+                // TODO: delete state.memory access after migration
+                state
+                    .memory
+                    .posts
+                    .get(post_id)
```
ulan: How do I check that this `get` will access the heap instead of the stable memory?

src/backend/env/post.rs
taggrx: I realized later that this whole test gets removed later in the second PR, as it tests more the allocator logic than anything else. So If you don't mind, I don't see a need to change here anything. Or would you feel more confident without compilation flags?

src/backend/env/post.rs
taggrx: I was thinking to do a round of tests with println's which would disclose the source of data. I even wanted to put this into PR but was afraid that it will blow up the change with debugging helpful for local testing _only_. I definitely can add it if you find it useful.

ulan: I might be missing something. Is the idea to migrate posts from stable memory to stable structures? Or do we load all cold posts into heap and then migrate?

src/backend/env/post.rs
ulan: If that specific test is not useful anymore, should we marked it as ignored? To make sure that we don't accidentally disable other important tests.

taggrx: Good idea, will add!

src/backend/env/post.rs
ulan: I don't see any special handling to make sure that the old stable memory doesn't overlap with the new stable memory for stable structure. So then my first question stands. This must not access stable memory because that would be broken, right?

taggrx: In `post_upgrade` we move everything into heap _and_ initialize structs. That's it.

The actual cold-archiving into the **new structs** happens then on am hourly timer (or using `archive` if needed).

So this particular `get` function will never be called before `post_upgrade` has finished. But after it, `state.memory.posts` struct's indices will all be empty as the cold-archiving will not touch it anymore and posts were moved out. So it cannot ever try to read the stable memory using the old allocator. Instead, we'll try the next line, which will already read from the new structs.

taggrx: @ulan yes, you are correct. What you seem to miss is that by moving posts from stable memory to the heap, the current allocator stops tracking the corresponding `post_id`. So if we moved all posts, the current allocator will not have any memory coordinates to resolve to to read anything. Does this make sense now?

ulan: Ok, if the function is either unreachable or a no-op, why do we have it here?

taggrx: I just tried to keep the change as small as possible to make it easier to review and reason about but it seems like I achieved the opposite 😂

ulan: In general I agree about minimizing changes to simplify review. This particular case relies on a subtle property. If that goes wrong for some reason, we will get corrupted data. I would remove this to make this case robust.

ulan: The reason is that as soon as we initialize stable structs, all accesses to the old allocator are potentially dangerous.

taggrx: Absolutely agree, I was not really aware of this subtle context in my head. Will remove the line!

taggrx: Removed.

chmllr (approved):
LGTM! I'll upgrade locally too.

chmllr:
Btw I could upgrade from main to this one and then to the finishing PR. Worked great. I also modified the parameter in code to make sure archived posts are restored and can be modified.

ulan:
I tested using the PocketIC (patched to print instructions and duration) and the production backup. The test was successful. Preliminary results.

- pre-upgrade uses 1.5B instructions
- post-upgrade uses 33B instructions
- 20 timer tasks each using 6B instructions.

users (query):
- before: instructions: 15389169, duration: 22ms
- after: instructions: 14934275, duration: 67ms

journal (query):
- before: instructions: 14575555, duration: 4ms
- after: instructions: 36301600, duration: 26ms

add_post (update):
- before: instructions: 8341867, duration: 6ms
- after: instructions: 516642, duration: 2ms

```
Execution of Method(Query("users")): instructions: 15389169, duration: 22ms
Execution of Method(Update("add_post")): instructions: 8341867, duration: 6ms
Execution of Method(Query("journal")): instructions: 14575555, duration: 4ms
Execution of Method(System(CanisterPreUpgrade)): instructions: 1538374729, duration: 300ms
Execution of Method(System(CanisterPostUpgrade)): instructions: 33796361478, duration: 3533ms
Execution of Method(Update("<ic-cdk internal> timer_executor")): instructions: 6646098035, duration: 1214ms
Execution of Method(Update("<ic-cdk internal> timer_executor")): instructions: 6892759552, duration: 1153ms
Execution of Method(Update("<ic-cdk internal> timer_executor")): instructions: 6653692450, duration: 1095ms
Execution of Method(Update("<ic-cdk internal> timer_executor")): instructions: 6622664707, duration: 1100ms
Execution of Method(Update("<ic-cdk internal> timer_executor")): instructions: 6688151193, duration: 1137ms
Execution of Method(Update("<ic-cdk internal> timer_executor")): instructions: 6609848420, duration: 1100ms
Execution of Method(Update("<ic-cdk internal> timer_executor")): instructions: 6721601388, duration: 1113ms
Execution of Method(Update("<ic-cdk internal> timer_executor")): instructions: 6581343760, duration: 1081ms
Execution of Method(Update("<ic-cdk internal> timer_executor")): instructions: 6649641357, duration: 1064ms
Execution of Method(Update("<ic-cdk internal> timer_executor")): instructions: 6567844629, duration: 1044ms
Execution of Method(Update("<ic-cdk internal> timer_executor")): instructions: 6533479058, duration: 1034ms
Execution of Method(Update("<ic-cdk internal> timer_executor")): instructions: 6551099225, duration: 1013ms
Execution of Method(Update("<ic-cdk internal> timer_executor")): instructions: 6491813183, duration: 1009ms
Execution of Method(Update("<ic-cdk internal> timer_executor")): instructions: 6579033855, duration: 993ms
Execution of Method(Update("<ic-cdk internal> timer_executor")): instructions: 6416037977, duration: 965ms
Execution of Method(Update("<ic-cdk internal> timer_executor")): instructions: 6453495951, duration: 976ms
Execution of Method(Update("<ic-cdk internal> timer_executor")): instructions: 6437505107, duration: 962ms
Execution of Method(Update("<ic-cdk internal> timer_executor")): instructions: 6430331391, duration: 954ms
Execution of Method(Update("<ic-cdk internal> timer_executor")): instructions: 6432385790, duration: 916ms
Execution of Method(Update("<ic-cdk internal> timer_executor")): instructions: 2283995965, duration: 432ms
Execution of Method(Query("users")): instructions: 14934275, duration: 67ms
Execution of Method(Query("journal")): instructions: 36301600, duration: 26ms
Execution of Method(Update("add_post")): instructions: 516642, duration: 2ms
```

ulan:
I need to run more tests to verify the performance results.

src/backend/env/mod.rs
```diff
@@ -194,7 +197,6 @@ pub struct State {
 
     migrations: BTreeSet<UserId>,
 
-    #[serde(default)]
```
leo-ship-it, line 197: Why we don't need this anymore?

src/backend/env/post.rs
```diff
@@ -91,6 +97,19 @@ impl PartialOrd for Post {
     }
 }
 
+impl Storable for Post {
+    const BOUND: ic_stable_structures::storable::Bound =
+        ic_stable_structures::storable::Bound::Unbounded;
```
leo-ship-it, line 102: Why don't we have an upper bound for post size? Wouldn't it be more efficient @ulan?

src/backend/env/post.rs
```diff
@@ -1,11 +1,17 @@
...
 use crate::reports::Report;
+use crate::{mutate, performance_counter};
+use ic_stable_structures::{StableBTreeMap, Storable};
 use serde::{Deserialize, Serialize};
+use std::borrow::Cow;
+use std::cell::RefCell;
+use std::cmp::{Ordering, PartialOrd};
+
+thread_local! {
+    pub static POSTS: RefCell<Option<StableBTreeMap<PostId, Post, crate::Memory>>> = Default::default();
```
leo-ship-it, line 13: Why don't we use a [StableLog](https://docs.rs/ic-stable-structures/latest/ic_stable_structures/log/struct.Log.html) instead of a map, we allocate the same amount of memory for each post, and to retrieve a post we just look at `postID * post size`?

I'm not really convinced about my suggestion I'm just challenging a bit the choices made.

leo-ship-it:
> I tested using the PocketIC (patched to print instructions and duration) and the production backup. The test was successful. Preliminary results.
> 
> * pre-upgrade uses 1.5B instructions
> * post-upgrade uses 33B instructions
> * 20 timer tasks each using 6B instructions.
> 
> users (query):
> 
> * before: instructions: 15389169, duration: 22ms
> * after: instructions: 14934275, duration: 67ms
> 
> journal (query):
> 
> * before: instructions: 14575555, duration: 4ms
> * after: instructions: 36301600, duration: 26ms
> 
> add_post (update):
> 
> * before: instructions: 8341867, duration: 6ms
> * after: instructions: 516642, duration: 2ms

To do a fair comparison shouldn't we look at cached queries vs not cached queries? Does it make a difference if the post/user is already in the heap or if it has to be loaded?

xqxpx:
> journal (query):
> 
> * before: instructions: 14575555, duration: 4ms
> * after: instructions: 36301600, duration: 26ms

I can't explain right away why the difference is so big given that my scanning of the entire stable memory has shown only a 15% regression... 🤔 

PS: could it be just a too big margin of error due to some variance in the stack?

src/backend/env/mod.rs
taggrx: This annotation is only needed for new fields. Then their values are persisted and deserialized from the stable memory.

src/backend/env/post.rs
taggrx: Because we have a cache above the stable memory. So posts need to be removed from the archive and put back eventually. I'll use the Log for ledger transaction, but posts are a volatile data. Or am I missing something about the Log structure?

src/backend/env/post.rs
taggrx: I am afraid due to posts sitting in hot memory first and only eventually are written to the stable memory, the question is how do we even ensure the posts didn't get e.g. too many reactions and can still fit into the bounded size? We don't want to serialize the post on every mutation to only count the seriazliation length, do we?

xqxpx:
Btw I also tested subsequent upgrades (after upgrading to the second PR) and made sure the stable memory is not growing. So the memory of Cell used for heap is allocated once and is used repeatedly because the heap fits there.

src/backend/env/post.rs
leo-ship-it: Right, when updating an entry in a stable btreemap do we need to serialize the whole post or just the field we want to update?

src/backend/env/post.rs
leo-ship-it: One thing that is still not clear to me is what we store in the buckets. I know that in the Taggr wasm we embed a bucket wasm and we store some stuff in them. Do we only store posts there? Or just pictures?

I also realized the Taggr transactions are not in stable logs, why don't we also add them to stable logs? On upgrade, we would just need to re-compute the balances by reading the transactions.

src/backend/env/post.rs
taggrx: Whole post.

src/backend/env/post.rs
taggrx: Buckets hold blobs only (images).

>I also realized the Taggr transactions are not in stable logs, why don't we also add them to stable logs? On upgrade, we would just need to re-compute the balances by reading the transactions.

The recomputation already happens right now, it's just all TXs are living in the heap. Changing this will be one of the next PRs.

xqxpx:
> I need to run more tests to verify the performance results.

Just so I understand you correctly: you would like to block on the release before you ran these tests, right? There's no rush, I just want to make sure we're aligned.

src/backend/env/post.rs
ulan: The posts have variable length, right? I thought that's why we don't specify upper bound so we avoid fragmentation.

ulan:
> > I need to run more tests to verify the performance results.
> 
> Just so I understand you correctly: you would like to block on the release before you ran these tests, right? There's no rush, I just want to make sure we're aligned.

Let me quickly run more tests. I wanted to double check the result. If I don't manage to finish that today, let's go ahead with the release. I think the change looks good.

src/backend/updates.rs
```diff
@@ -41,7 +42,13 @@ fn init() {
 
 #[pre_upgrade]
 fn pre_upgrade() {
-    mutate(env::memory::heap_to_stable)
+    HEAP.with(|cell| {
```
ulan, line 46: FYI: this is currently failing in standalone PocketIC tests that start with an empty canister because the stable structs are not initialized.  Would it make sense to initialize them from `init` as well?

ulan (approved):

src/backend/updates.rs
taggrx: Strange! Usually the e2e tests fail as well if something is wrong with unintiailized canister... I'll check this.

taggrx: I figured: this is `pre_upgrade` and I disabled upgrade tests!

PR #22, opened 2024-02-24
Bugfixes

xqxpx:
- realm gating
- styling
- content filtering


leo-ship-it (approved):

src/backend/queries.rs
```diff
@@ -387,13 +387,27 @@ fn hot_realm_posts() {
     });
 }
 
+fn personal_filter(state: &State, user: Option<&User>, post: &Post) -> bool {
```
leo-ship-it, line 390: nit: I would name it `should_filter_personal`

PR #23, opened 2024-02-26
Block spaced allocation

xqxpx:


src/backend/updates.rs
```diff
@@ -56,6 +57,10 @@ fn post_upgrade() {
     }
     stable_to_heap_core();
     mutate(|state| state.load());
+
+    // TODO: DELETE BEFORE THE NEXT RELEASE
+    mutate(|state| state.memory.set_block_size(1));
```
ulan, line 62: Shouldn't this be the first thing in the `post_upgrade` to ensure that the allocator never sees the block size of 0.

src/backend/env/memory.rs
```diff
@@ -179,7 +197,12 @@ impl Default for Allocator {
 }
 
 impl Allocator {
-    fn alloc(&mut self, n: u64) -> Result<u64, String> {
+    fn get_allocation_length(&self, n: u64) -> u64 {
+        (n + self.block_size_bytes - 1) / self.block_size_bytes * self.block_size_bytes
```
ulan: Wdyt about this to make it more robust against accidental 0:
```
let block_size = self.block_size_bytes.max(1);
(n + block_size - 1) / block_size * block_size
```

src/backend/updates.rs
ulan: Ah, nevermind. This is the first opportunity of the loading the state.

src/backend/updates.rs
```diff
@@ -64,13 +69,40 @@ fn post_upgrade() {
...
+            Post::mutate(state, &post_id, |post| {
+                post.archived = false;
+                Ok(())
+            })
+            .unwrap();
+        }
+
+        // The noop mutation above should move all posts to the heap
+        assert_eq!(state.memory.posts.len(), 0);
+        assert_eq!(state.posts.len(), num_posts);
```
ulan, line 101: Can we also assert that allocator consists of a single segment? That should hold at this point IIUC.

src/backend/env/memory.rs
```diff
@@ -214,7 +237,8 @@ impl Allocator {
         Ok(start)
     }
 
-    fn free(&mut self, offset: u64, size: u64) -> Result<(), String> {
+    fn free(&mut self, offset: u64, len: u64) -> Result<(), String> {
+        let size = self.get_allocation_length(len);
```
ulan, line 246: If block size changes between `alloc` and `free`, then this will lead to data corruption, so we need to be careful when changing the block size.

ulan (approved):
Looks good! Let me run a few tests before the release.

ulan (commented):
Edit: this was a problem in the test itself, which used the old definition of `check`. The performance is the same as before, there is no regression.

> Something is off. My test is showing at least 100x slowdown in terms of instructions and time (running out of instructions in `check`). Investigating.

ulan:
There is no regression in number of instructions, but duration regresses. Investigating.

```
Execution of Method(Query("check")): instructions: 343101952, duration: 45ms
Execution of Method(Query("check")): instructions: 361376047, duration: 145ms

Execution of Method(Query("users")): instructions: 16256770, duration: 24ms
Execution of Method(Query("users")): instructions: 14429103, duration: 65ms

Execution of Method(Query("journal")): instructions: 15412744, duration: 6ms
Execution of Method(Query("journal")): instructions: 15603290, duration: 18ms

Execution of Method(Update("add_post")): instructions: 9011414, duration: 12ms
Execution of Method(Update("add_post")): instructions: 393258, duration: 2ms
- 
```

ulan:
Looks like the longer durations are side-effects of reallocation. Here are results of replacing 300 with 1:
```
Execution of Method(Query("check")): instructions: 361355703, duration: 145ms
Execution of Method(Query("users")): instructions: 14434217, duration: 64ms
Execution of Method(Query("journal")): instructions: 15603565, duration: 18ms
Execution of Method(Update("add_post")): instructions: 404752, duration: 2ms
```

ulan:
Since durations are strange and I am not sure if that's real or artifact of the testing framework, let's not block on them.

I checked that upgrading once work. We also need to check that upgrading twice (to the next version with adjusted post-upgrade) also works.

Note that with 10K posts per archiving, it will take 60 hours to archive all posts.

ulan:
I tested upgrading twice and three times to Wasm binary that has the post-upgrade code removed. I am observing doubling of the stable memory after each upgrade.

@taggrx : could you please check if upgrading twice keeps the stable memory size the same for you? Maybe I missed some code.

src/backend/env/memory.rs
taggrx: Exactly, this is why we need to unload everything in post_upgrade first.

I plan to comment out the block changing logic after this upgrade as well.

xqxpx:
> I tested upgrading twice and three times to Wasm binary that has the post-upgrade code removed. I am observing doubling of the stable memory after each upgrade.
> 
> @taggrx : could you please check if upgrading twice keeps the stable memory size the same for you? Maybe I missed some code.

Hey Ulan, if you do not archive the posts after the upgrade, then the heap (containing all data) is written behind what the allocator thinks is the boundary. So in this situation the first upgrade would indeed allocate this additional space. But any subsequent upgrade should not double it and I indeed cannot reproduce this behavior.

Is it possible that you forgot to remove [this line](https://github.com/TaggrNetwork/Taggr/pull/23/files#diff-b5aec5daefc97057045961cfcf63573875fe6e693091935078952e5e90467a77R61)?

xqxpx:
@ulan I just added your suggestions (thanks a lot) and a new way of still having a backup. It's a tiny commit, please check when you have time, but I'll speculatively create the release in the mean time.

src/backend/updates.rs
```diff
@@ -672,6 +711,16 @@ fn check_candid_interface_compatibility() {
     );
 }
 
+#[update]
+fn backup() {
```
ulan, line 715: A malicious stalwart can drain cycles and DoS the canister by calling `archive` and `backup` in a loop. Is this a risk we are willing to accept?

ulan:
> > I tested upgrading twice and three times to Wasm binary that has the post-upgrade code removed. I am observing doubling of the stable memory after each upgrade.
> > @taggrx : could you please check if upgrading twice keeps the stable memory size the same for you? Maybe I missed some code.
> 
> Hey Ulan, if you do not archive the posts after the upgrade, then the heap (containing all data) is written behind what the allocator thinks is the boundary. So in this situation the first upgrade would indeed allocate this additional space. But any subsequent upgrade should not double it and I indeed cannot reproduce this behavior.
> 
> Is it possible that you forgot to remove [this line](https://github.com/TaggrNetwork/Taggr/pull/23/files#diff-b5aec5daefc97057045961cfcf63573875fe6e693091935078952e5e90467a77R61)?

To close the loop: this was an issue with my testing setup. I redid the test and it looks good.

PR #25, opened 2024-02-29
Release

xqxpx:
- post-release clean up
- adds an index for root posts
- decreases user data in the heap
- excludes comments from user tags tab
- reset the report from user Zikhas

### DIFF
--- a/src/backend/env/mod.rs
+++ b/src/backend/env/mod.rs
@@ -584,7 +584,7 @@ impl State {
         filter: Option<&dyn Fn(&Post) -> bool>,
     ) -> Box<dyn Iterator<Item = &'_ Post> + '_> {
         let mut hot_posts = self
-            .last_posts(realm, offset, time().saturating_sub(4 * WEEK), false)
+            .last_posts(realm, offset, 0, false)
             .filter(|post| !matches!(post.extension, Some(Extension::Proposal(_))))
             .filter(|post| filter.map(|f| f(post)).unwrap_or(true))
             .take(1000)
--- a/src/backend/env/mod.rs
+++ b/src/backend/env/mod.rs
@@ -182,7 +182,12 @@ pub struct State {
 
     pub last_nns_proposal: u64,
 
+    // TODO: delete
+    #[serde(skip)]
     pub root_posts: usize,
+
+    #[serde(default)]
+    pub root_posts_index: Vec<PostId>,
 
     e8s_for_one_xdr: u64,
 
@@ -1982,18 +1987,28 @@ impl State {
                         .copied(),
                 ),
                 _ => {
-                    let last_id = if offset > 0 {
-                        offset
+                    if with_comments {
+                        let last_id = if offset > 0 {
+                            offset
+                        } else {
+                            self.next_post_id
+                        };
+                        Box::new((0..last_id).rev())
                     } else {
-                        self.next_post_id
-                    };
-                    Box::new((0..last_id).rev())
+                        Box::new(
+                            self.root_posts_index
+                                .iter()
+                                .rev()
+                                .skip_while(move |post_id| offset > 0 && *post_id > &offset)
+                                .copied(),
+                        )
+                    }
                 }
             };
         Box::new(
             iter.filter_map(move |i| Post::get(self, &i))
                 .take_while(move |post| post.creation_timestamp() >= genesis)
-                .filter(move |post| !post.is_deleted() && (with_comments || post.parent.is_none())),
+                .filter(move |post| !post.is_deleted()),
         )
     }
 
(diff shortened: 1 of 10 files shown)


ulan:
I'll try to take a look today after work or tomorrow.

chmllr (approved):
LGTM

xqxpx:
I added a couple of minor fixes.

src/backend/env/mod.rs
```diff
@@ -1554,19 +1561,36 @@ impl State {
             )
         });
 
+        let log_time = |state: &mut State, frequency| {
+            state.logger.debug(format!(
+                "{} routine finished after `{}` ms.",
```
ulan, line 1566: FYI: this works on granularity of rounds (if there are async calls before). Otherwise the result will be 0.

src/frontend/src/post.tsx
```diff
@@ -122,10 +122,18 @@ export const PostView = ({
 
     React.useEffect(() => {
         const article: any = refArticle.current;
-        if (!prime && article && article.scrollHeight > article.clientHeight) {
-            article.classList.add("overflowing");
-            setShowExpandButton(true);
-        }
+        // We delay shortening the post by half a second in the hope that images would
+        // have loaded by that time. But it should be improved in the future.
+        setTimeout(() => {
```
ulan: Could it cause the annoying effect of moving re-arranging posts while the user is looking at them? If so, then I am not sure this is a good change, since it gives an impression of a slow website.

src/backend/updates.rs
```diff
@@ -69,46 +65,31 @@ fn post_upgrade() {
...
-        state.memory.set_block_size(300);
-
-        ic_cdk::println!(
-            "all posts loaded into heap (instructions: {}B, current SM size: {}Mb)",
-            performance_counter(0) / 1000000000,
-            (ic_cdk::api::stable::stable_size() << 16) / 1024 / 1024
-        );
+async fn async_post_upgrade_fixtures() {
+    // Fill the root posts index.
+    mutate(|state| {
```
ulan, line 83: Did you verify that this completes in 20B instructions?

ulan (approved):
Lgtm modulo the UI timeout change (see question).

src/backend/env/mod.rs
taggrx: You're saying this is not the wall time, right? I expect that. I would be still curious to see what we get. Because I have no explanation for this: https://6qfxa-ryaaa-aaaai-qbhsq-cai.icp0.io/#/post/717170

src/frontend/src/post.tsx
taggrx: Yes, it can fore the upper posts. However, this is already happening when images get loaded because they are stretched and Taggr does not know in advance the exact rendered height. Also I am not sure what's worse - cutting the post in a wrong way, or what you describe? :D

src/backend/updates.rs
taggrx: I measured on the local replica and since this query is similar to `check` which seems to pass even as a simply query, I was not worried about that in the upgrade. Moreover, I just added a new commit with the counter of instructions that will always print how much was spent.

The counter says the post-upgrade spent 31B instructions. I am confused. The budget for upgrades is 20? Why does it work on a local replica?

src/frontend/src/post.tsx
taggrx: Btw, two more points:

- The effects you are afraid of will only happen on certain posts, not on all of them (those which have images and they are too long). These posts have the effect you describe already as I described above. 
- There is one way I wanted to experiment with: propagating a callback all the way to images which they would call notifying about a load. But this could be a much heavier change which I wanted to try later (as stated in comments).

taggrx: I thought a bit more and decided not to put this change into the release, because it can be see as controversial.

src/backend/updates.rs
taggrx: Btw, are these budgets in any kind of official spec? Or where could I look them up?

taggrx: Ok I found this: https://6qfxa-ryaaa-aaaai-qbhsq-cai.icp0.io/#/post/472562 Now it makes sense.

I keep forgetting what the limits are: 2/20 or 20/200...

ulan: Yeah, it is 5B (query) / 20B (update) / 200B (upgrade):
https://internetcomputer.org/docs/current/developer-docs/smart-contracts/maintain/resource-limits

src/frontend/src/post.tsx
ulan: Postponing sgtm. Let's experiment more here. Maybe we find a better way to improve it.

src/backend/env/mod.rs
ulan: Replied on that thread. The change sound good for that purpose.

taggrx: I will just set it to current time as you suggested. The reason I add 1h is to not have it drifting...

taggrx: Ok, maybe in the next release.

PR #26, opened 2024-02-29
Remove data pre-loading

xqxpx:
This change removes preloading of backend data (like all user names and some data of active realms) and loads this data on demand. This should remove the loading of hundreds of kilobytes of data when opening Taggr (and every 10m).

To compensate for the data missing on UI, posts are now extended with meta information and each user profile has a vector of realm ids they control.

### DIFF
--- a/src/backend/env/mod.rs
+++ b/src/backend/env/mod.rs
@@ -628,7 +628,7 @@ impl State {
     pub fn edit_realm(
         &mut self,
         principal: Principal,
-        name: String,
+        realm_id: String,
         realm: Realm,
     ) -> Result<(), String> {
         let Realm {
@@ -646,7 +646,7 @@ impl State {
         let user = self.principal_to_user(principal).ok_or("no user found")?;
         let user_id = user.id;
         let user_name = user.name.clone();
-        let realm = self.realms.get_mut(&name).ok_or("no realm found")?;
+        let realm = self.realms.get_mut(&realm_id).ok_or("no realm found")?;
         if !realm.controllers.contains(&user_id) {
             return Err("not authorized".into());
         }
@@ -659,9 +659,21 @@ impl State {
         let description_change = realm.description != description;
         realm.description = description;
         if realm.controllers != controllers {
+            let mut old_names = Vec::default();
+            let mut new_names = Vec::default();
+            for user_id in &realm.controllers {
+                let controller = self.users.get_mut(&user_id).expect("no user found");
+                controller.controlled_realms.remove(&realm_id);
+                old_names.push(controller.name.clone());
+            }
+            for user_id in &controllers {
+                let controller = self.users.get_mut(&user_id).expect("no user found");
+                controller.controlled_realms.insert(realm_id.clone());
+                new_names.push(controller.name.clone());
+            }
             self.logger.info(format!(
                 "Realm /{} controller list was changed from {:?} to {:?}",
-                name, &realm.controllers, &controllers
+                realm_id, old_names, new_names
             ));
         }
         realm.controllers = controllers;
@@ -674,10 +686,10 @@ impl State {
         realm.adult_content = adult_content;
         if description_change {
             self.notify_with_filter(
-                &|user| user.realms.contains(&name),
+                &|user| user.realms.contains(&realm_id),
                 format!(
                     "@{} changed the description of the realm /{}! ",
-                    user_name, name
+                    user_name, realm_id
                 ) + "Please read the new description to avoid potential penalties for rules violation!",
             );
         }
--- a/src/backend/env/mod.rs
+++ b/src/backend/env/mod.rs
@@ -662,12 +662,12 @@ impl State {
             let mut old_names = Vec::default();
             let mut new_names = Vec::default();
             for user_id in &realm.controllers {
-                let controller = self.users.get_mut(&user_id).expect("no user found");
+                let controller = self.users.get_mut(user_id).expect("no user found");
                 controller.controlled_realms.remove(&realm_id);
                 old_names.push(controller.name.clone());
             }
             for user_id in &controllers {
-                let controller = self.users.get_mut(&user_id).expect("no user found");
+                let controller = self.users.get_mut(user_id).expect("no user found");
                 controller.controlled_realms.insert(realm_id.clone());
                 new_names.push(controller.name.clone());
             }
--- a/src/backend/env/post.rs
+++ b/src/backend/env/post.rs
@@ -143,15 +143,14 @@ impl Post {
     /// Returns the post with some meta information needed for by the UI.
     pub fn with_meta<'a>(&'a self, state: &'a State) -> (&'_ Self, Meta<'_>) {
         let user = state.users.get(&self.user).expect("no user found");
+        let realm = self
+            .realm
+            .as_ref()
+            .and_then(|realm_id| state.realms.get(realm_id));
         let meta = Meta {
             author_name: user.name.as_str(),
             author_rewards: user.rewards(),
-            realm_color: self.realm.as_ref().and_then(|realm_id| {
-                state
-                    .realms
-                    .get(realm_id)
-                    .map(|realm| realm.label_color.as_str())
-            }),
+            realm_color: realm.map(|realm| realm.label_color.as_str()),
         };
         (self, meta)
     }
--- a/src/backend/env/post.rs
+++ b/src/backend/env/post.rs
@@ -50,6 +50,13 @@ pub enum Extension {
 }
 
 #[derive(Clone, Default, Serialize, Deserialize)]
+pub struct Meta<'a> {
+    author_name: &'a str,
+    author_rewards: i64,
+    realm_color: Option<&'a str>,
+}
+
+#[derive(Clone, Default, Serialize, Deserialize)]
 pub struct Post {
     pub id: PostId,
     pub body: String,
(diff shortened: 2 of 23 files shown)


ulan:
I guess it is not ready for review yet since it is marked as draft?

src/backend/env/mod.rs
```diff
@@ -659,9 +659,21 @@ impl State {
...
+            }
+            for user_id in &controllers {
+                let controller = self.users.get_mut(user_id).expect("no user found");
+                controller.controlled_realms.insert(realm_id.clone());
+                new_names.push(controller.name.clone());
+            }
             self.logger.info(format!(
                 "Realm /{} controller list was changed from {:?} to {:?}",
-                name, &realm.controllers, &controllers
+                realm_id, old_names, new_names
```
chmllr, line 665: nice!

src/backend/env/post.rs
```diff
@@ -133,6 +143,26 @@ impl Post {
         }
     }
 
+    /// Returns the post with some meta information needed for by the UI.
```
chmllr:
```suggestion
    /// Returns the post with some meta information needed by the UI.
```

chmllr (approved):
LGTM, but UI changes is difficult to evaluate. It would be good to check on staging.

src/backend/env/post.rs
taggrx: Thanks Christian!

src/backend/env/post.rs
```diff
@@ -133,6 +143,26 @@ impl Post {
         }
     }
 
+    /// Returns the post with some meta information needed by the UI.
+    pub fn with_meta<'a>(&'a self, state: &'a State) -> (&'_ Self, Meta<'_>) {
```
ulan, line 147: Do you know how much it is going to increase the size of a Post. I wonder if we are far away from the 3MB max response size limit in all queries that return posts.

ulan:
I wonder how many posts Taggr serves per day.

Let's say we have `A` active users. The number of total users is `U`.  An active user loads `P` posts per ten minutes. Then we are comparing the following in terms of traffic:

- `A * P * sizeof(meta)` vs `A * U * sizeof(meta)`.

I guess `U >> P`, so this is probably a good change.

Another good point is that the existing queries that worked in `O(U)` were not scalable and would break at some point due to the 3MB query response limit.

src/backend/env/post.rs
taggrx: No, actually we never ever serve more than 30 posts because everything is paged. Where we bump into limits is when we _search_ for posts to fill a page. So these meta augmentation will be only ever called on 30 posts in the same query.

xqxpx:
> Another good point is that the existing queries that worked in `O(U)` were not scalable and would break at some point due to the 3MB query response limit.

That's where we have no room for negotiation with IC, I'm afraid 😅 

Then there is another ugly side effect of the current solution: when a user registered and writes a post in-between of 10m backend data fetches, users see a post from an "N/A" and are very confused.

With this change, the displayed data will be always complete.

ulan:
After this ships. Maybe we could follow up with an optimization where a query collects unique users and realms in the set of posts and returns `(posts, user_data, realm_data)`. I guess in most cases there will be duplicate users and realms.

src/frontend/src/common.tsx
```diff
@@ -542,12 +564,25 @@ export const UserLink = ({
...
     classNameArg,
     profile,
+    name,
 }: {
     id: UserId;
     classNameArg?: string;
     profile?: boolean;
+    name?: string;
 }) => {
-    const userName = window.backendCache.users[id];
```
ulan, line 550: I wonder if it would make sense to keep this cache and populate it from post metadata. The cache would be update only if the user is missing from it.

taggrx: That's an interesting idea, but what I'd do first is to measure the hit/miss ratio: I am afraid that in many cases where we need to load say 20 user names (e.g. by opening the post info section with reactions, watchers, tippers and so on), we still would have a small number of misses because we never loaded posts from those users and would still need to do a query. But in that case the latency would be already not zero...

ulan: It would be useful when the user returns to the same posts / comments / reactions multiple times.

src/frontend/src/common.tsx
```diff
@@ -560,8 +595,42 @@ export const UserLink = ({
...
+            (await window.api.query<UserData>("users_data", removeNAs(ids))) ||
+                {},
+        );
+
+    React.useEffect(() => {
+        if (removeNAs(ids) && Object.entries(loadedNames).length == 0)
+            loadNames();
+    }, []);
+
+    return Object.keys(names).length == 0 && removeNAs(ids).length > 0 ? (
```
ulan: Nit: this check is inconsistent with the similar check in line 620.

ulan:
I have concerns about the responsiveness of the UI with the stateful `UserLink` / `UserList` components. I wonder if more incremental approach would be better?

- change the query that returns all users to return the most recent 100-500 active users.
- keep the cache and update it every 10 minutes as before.
- if the user is in the cache the use the stateless `UserLink` otherwise use `LoadUserLink` that loads the user and saves it the cache.

Wdyt?

src/frontend/src/common.tsx
taggrx: Thanks, yes, this could be improved to not load an empty list.

xqxpx:
> I have concerns about the responsiveness of the UI with the stateful `UserLink` / `UserList` components. I wonder if more incremental approach would be better?
> 
> * change the query that returns all users to return the most recent 100-500 active users.
> * keep the cache and update it every 10 minutes as before.
> * if the user is in the cache the use the stateless `UserLink` otherwise use `LoadUserLink` that loads the user and saves it the cache.
> 
> Wdyt?

But what's your concern based on? I tested this change with a 3g throttling emulation on mainnet data. One noticable change is that when you open the post info section, it takes longer than before to display the section with the emoji donors. Same with polls and proposals: e.g. voter names take a bit of time to show up. But this is so minor, that it's hardly a blocker. All posts take the names from the metas right?

xqxpx:
@ulan  Btw, I am absolutely not against caching, but maybe in a next step with some smart pre-loading.

I just deployed this change to staging, feel free to check there.

ulan:
> > I have concerns about the responsiveness of the UI with the stateful `UserLink` / `UserList` components. I wonder if more incremental approach would be better?
> > 
> > * change the query that returns all users to return the most recent 100-500 active users.
> > * keep the cache and update it every 10 minutes as before.
> > * if the user is in the cache the use the stateless `UserLink` otherwise use `LoadUserLink` that loads the user and saves it the cache.
> > 
> > Wdyt?
> 
> But what's your concern based on? I tested this change with a 3g throttling emulation on mainnet data. One noticable change is that when you open the post info section, it takes longer than before to display the section with the emoji donors. Same with polls and proposals: e.g. voter names take a bit of time to show up. But this is so minor, that it's hardly a blocker. All posts take the names from the metas right?

I am not sure how this affects the rendering performance overall. Intuitively it seems that it should be more difficult for React to render, but maybe it is not the case. If it is the common operations to expand the post to see the reactions, then that delay will perceived as slowness by the users.

ulan:
> @ulan Btw, I am absolutely not against caching, but maybe in a next step with some smart pre-loading.
> 
> I just deployed this change to staging, feel free to check there.

There is about 500ms - 1s lag before the watcher list appears and it also changes the layout.

xqxpx:
> then that delay will perceived as slowness by the users.

Maybe there will be some degradation here and there, but you only need to open some other IC dapps to understand this is nit picking. I think minor degradations sometime are ok if they can be fixed quickly and this happened very often in the past as well. OTOH I can't count cases where some minor and cheap improvement worked so good, that I didn't touch it for months. I strongly believe that faster and simpler iterations (especially for non critical changes) is a much better way to work on a side project. Because when I would start pursuing perfection in every aspect and the granularity of work increases I won't be able to hold the scope of the entire change and start introducing more bugs, as I can only work on things in short but very intense periods of time. I have no actual time for any kind of deep work, unfortunately.

Please trust a bit more - it worked quite well in the past 2 years. (apart from uploading a wrong binary once 😂 but this a different issue)

xqxpx:
> > @ulan Btw, I am absolutely not against caching, but maybe in a next step with some smart pre-loading.
> > I just deployed this change to staging, feel free to check there.
> 
> There is about 500ms - 1s lag before the watcher list appears and it also changes the layout.

But this is by far not blocking right? Or would you still prefer to have a cache in the same release?

ulan:
Would you be fine with merging this after the release and re-introducing the cache before the next release?

> you only need to open some other IC dapps to understand this is nit picking.

That's why I don't use other IC dapps, but use Taggr. At least for me the responsiveness of the UI is a critical feature. I am not trying to nitpick or push for perfection. To me it seems more incremental and smaller change to keep the cache and populate it either from the post meta data or from the active users.

> Please trust a bit more - it worked quite well in the past 2 years.

How many users were there in the first 2 years? When is it the good time to start treating it as an actual production project?

xqxpx:
> Would you be fine with merging this after the release and re-introducing the cache before the next release?

Not sure I understand the question: if we release this change in the next upgrade, then I'd definitely introduce the cache before the next release. The cache, as a stand-alone change would be simple enough, I only don't like pushing more complexity into _this_ PR if it will be part of the next release.

> How many users were there in the first 2 years? When is it the good time to start treating it as an actual production project?

For me it's the income from the project, not the number of users that define the level of maturity and commitment I can sigh up for. The reality is that if I still cannot afford working full time on it even at 100k users, then it's still just a side-project actually and I can work on it here and there when I get the time...

Btw, I would be absolutely fine to merge this PR later and not including in the release. So that the next release contains this change and the cache.

ulan:
> Btw, I would be absolutely fine to merge this PR later and not including in the release. So that the next release contains this change and the cache.

Yep, that's what I meant. Maybe there was misunderstanding. As long as this is not released without the cache, it is looking good.

ulan (approved):

xqxpx:
> > Btw, I would be absolutely fine to merge this PR later and not including in the release. So that the next release contains this change and the cache.
> 
> Yep, that's what I meant. Maybe there was misunderstanding. As long as this is not released without the cache, it is looking good.

Ah I see, yes let's skip this and not include in the release. I'll merge later.

PR #27, opened 2024-03-02
feat: make minting and icp rewards mutually exclusive

xqxpx:
Implementation of https://taggr.wtf/#/post/670617:

- Every user can switch to the mining mode
- In mining mode no rewards for reactions can be received and paid out
- Only in mining mode can a user mine new TAGGR.
- No cycles top up is possible anymore.

### DIFF
--- a/src/backend/env/mod.rs
+++ b/src/backend/env/mod.rs
@@ -1466,7 +1466,7 @@ impl State {
                     });
                 };
             }
-            mutate(|state| state.pending_nns_proposals.remove(&post_id));
+            mutate(|state| state.pending_nns_proposals.remove(&proposal_id));
         }
 
         // fetch new proposals
@@ -1602,6 +1602,12 @@ impl State {
             State::daily_chores(now).await;
             mutate(|state| {
                 state.last_daily_chores += DAY;
+                state.logger.debug(format!(
+                    "Pending NNS proposals: `{}`, pending polls: `{}`, migrations: `{}`.",
+                    state.pending_nns_proposals.len(),
+                    state.pending_polls.len(),
+                    state.migrations.len(),
+                ));
                 log_time(state, "Daily");
             });
         }
--- a/src/backend/env/mod.rs
+++ b/src/backend/env/mod.rs
@@ -1083,13 +1083,21 @@ impl State {
             .users
             .values()
             .flat_map(|user| user.mintable_tokens(self, user_shares, boostraping_mode))
-            .filter(|(_, balance)| *balance > 0)
         {
             tokens_to_mint
                 .entry(user_id)
                 .and_modify(|balance| *balance += tokens)
                 .or_insert(tokens);
         }
+
+        tokens_to_mint.retain(|user_id, balance| {
+            *balance > 0
+                && self
+                    .users
+                    .get(user_id)
+                    .map(|user| user.eligible_for_minting())
+                    .unwrap_or_default()
+        });
 
         tokens_to_mint
     }
@@ -1215,12 +1223,22 @@ impl State {
     }
 
     pub fn collect_new_rewards(&mut self) -> HashMap<UserId, u64> {
-        self.users
-            .values_mut()
-            .filter_map(|u| {
-                (u.rewards() > 0).then_some((u.id, u.take_positive_rewards() as Credits))
-            })
-            .collect()
+        let mut payouts = HashMap::default();
+
+        for user in self.users.values_mut() {
+            let rewards = user.take_positive_rewards();
+            if rewards == 0 {
+                continue;
+            };
+            // All miner rewards are burned.
+            if user.miner {
+                self.burned_cycles += rewards;
+            } else {
+                payouts.insert(user.id, rewards as Credits);
+            }
+        }
+
+        payouts
     }
 
     async fn distribute_icp(
@@ -3047,6 +3065,7 @@ pub(crate) mod tests {
             vec![],
             Default::default(),
             false,
+            false,
             false
         )
         .is_err());
@@ -3058,6 +3077,7 @@ pub(crate) mod tests {
             Default::default(),
             vec![],
             Default::default(),
+            false,
             false,
             false
         )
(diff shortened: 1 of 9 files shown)


src/frontend/src/settings.tsx
```diff
@@ -148,6 +151,17 @@ export const Settings = ({ invite }: { invite?: string }) => {
                 />
                 {user && (
                     <>
+                        <div className="bottom_half_spaced">
+                            Participate in token minting
```
ulan: Wdyt being more descriptive about the options?

```
Minting and rewards:
- Mint TAGGR tokens
- Receive ICP rewards
```

The underlying `miner` flag can remain the same as before.

src/backend/env/mod.rs
```diff
@@ -1083,14 +1083,22 @@ impl State {
             .users
             .values()
             .flat_map(|user| user.mintable_tokens(self, user_shares, boostraping_mode))
-            .filter(|(_, balance)| *balance > 0)
```
ulan, line 1086: I think it was a good optimization to avoid creating a new entry in the hash table for users that don't get any tokens.

Maybe even better optimization would be to adjust `mintable_tokens` to never return such entries.

src/backend/env/mod.rs
```diff
@@ -1215,12 +1223,22 @@ impl State {
...
+        let mut payouts = HashMap::default();
+
+        for user in self.users.values_mut() {
+            let rewards = user.take_positive_rewards();
+            if rewards == 0 {
+                continue;
+            };
+            // All miner rewards are burned.
+            if user.miner {
+                self.burned_cycles += rewards;
```
ulan, line 1233: Just to double check: this will account these rewards as revenue, so they will be distributed to users as revenue?

ulan (approved):
Lgtm!

src/frontend/src/settings.tsx
```diff
@@ -148,6 +151,17 @@ export const Settings = ({ invite }: { invite?: string }) => {
...
+                        </div>
+                        <select
+                            value={miner}
+                            className="bottom_spaced"
+                            onChange={(event) => setMiner(event.target.value)}
+                        >
+                            <option value="true">YES</option>
+                            <option value="false">NO</option>
```
aodl: Rather than a binary choice, have you considered a sliding scale? i.e. how much of a users TAGGR holding should be directed towards revenue distribution, and how much should be directed towards mining?

src/backend/env/mod.rs
taggrx: Yes!

src/frontend/src/settings.tsx
taggrx: No, I didn't.

src/frontend/src/settings.tsx
taggrx: Good idea!

src/backend/env/mod.rs
taggrx: Done!

src/frontend/src/settings.tsx
taggrx: To clarify: this should be discussed on Taggr of course, not in code reviews :)

aodl: Appologies for the late response, I was investigating the potential for undesirable implications that I think a binary selection presents.

Restricting this to a binary choice strongly incentivises each user (particularly wales) to create a dummy TAGGR account alongside their main TAGGR account. 

It would be within a user's interest to set one account to mining and the other account to revenuing sharing (instead of mining). The revenue sharing account would be the dummy account that isn't required to exhibit much activity to benefit from revenue distribution (it's based on TAGGR holding). The mining account would be the users main account where they engage with the community and receive rewards to mint. A savy user may in fact make sure that the TAGGR balance in their main account remains low enough to benefit from maximum [priority_factor](https://github.com/TaggrNetwork/Taggr/blob/3637099ae4423e38a36897b7ed8d3bd21b79c596/src/backend/env/user.rs#L563), while increasing the TAGGR balance in their dummy account to maximise their revenue share.

Instead of a proliferation of main and dummy accounts (which would clutter up the community and reduce transparity), it would be more transparent to allow a single account to devote a porition of their TAGGR holding to mining, and the rest to revenue distribution (a sliding scale rather than a binary choice). This ensures that sneaky behaviour is not rewarded, and genuine users with integrity aren't at risk of missing out.

taggrx: > Restricting this to a binary choice strongly incentivises each user (particularly wales) to create a dummy TAGGR account alongside their main TAGGR account.

It does not help with anything. ICP spent on rewards of the "main account" will not be returned to this user.

taggrx: > Restricting this to a binary choice strongly incentivises each user (particularly wales) to create a dummy TAGGR account alongside their main TAGGR account.

It does not help with anything. ICP spent on rewards of the "main account" will not be returned to this user.

aodl: The algorithm for distributing revenue is very different to the algorithm for distributing minted TAGGR. They reward different things, and therefore it is within a users interest to split their TAGGR into two separate accounts to game the system by appealing to each algorithm separately (given that this proposal completely decouples them)
- TAGGR minting prefers TAGGR balances that aren't too high (in order to have a higher priority factor). It also prefers an active account that generates engagement
- ICP revenue distribution prefers TAGGR balances that are high (by contrast to minting). It's not fussed about how much engagement you're getting (so this account would be easy to run)

aodl: Forcing users to confront the decision to mine TAGGR or share in ICP distribution makes this exploit significantly more obvious. It also makes it seem easier to achieve (given that there would be no incentive to drive any engagement on the ICP distribution account).

What are your thoughts about introducing a sliding scale, instead of a binary choice (thereby removing the incentive to create multiple accounts)? It doesn't seem like it would introduce much complexity, and could potentially open up other options in the future (such as staking/locking up a portion of your TAGGR, providing other benefits)

aodl: https://taggr.link/#/post/775804

PR #28, opened 2024-03-05
Release

xqxpx:
- Fix: display speculative revenue on the dashboard (like the speculative token mint but for ICP)
- Improvement: Display VP on all options of a poll
- Feature: cap donatable tokens by the minted account according to the [DAO poll](https://6qfxa-ryaaa-aaaai-qbhsq-cai.icp0.io/#/post/803089).
- Feature: introduce a difficulty adjustment factor that will make the minting target `11` years according to the result of the [DAO poll](https://6qfxa-ryaaa-aaaai-qbhsq-cai.icp0.io/#/thread/835836) computed by VP-weighting. Attention: this is a temporary solution for delaying the minting until we have figured out a more elegant model doing what we want.
- Fix: top 100 holders chart
- Fix: paging on landing leading to too short streams

### DIFF
--- a/src/backend/env/config.rs
+++ b/src/backend/env/config.rs
@@ -23,6 +23,11 @@ pub struct Config {
     pub token_symbol: &'static str,
     pub maximum_supply: Token,
 
+    // This is a custom difficulty adjustment that can be set by the DAO in emergency cases
+    // when the speed of minting does not meet DAO's expectations. Currently, we consider this as a
+    // temporary measure to slow down the minting until the DAO agrees on a better approach.
+    pub difficulty_amplification: u64,
+
     pub max_age_hot_post_days: u64,
 
     pub downvote_counting_period_days: u64,
@@ -186,6 +191,11 @@ pub const CONFIG: &Config = &Config {
     minting_threshold_percentage: 5,
 
     active_user_share_for_minting_promille: 10,
+
+    #[cfg(any(test, feature = "dev"))]
+    difficulty_amplification: 1,
+    #[cfg(not(any(test, feature = "dev")))]
+    difficulty_amplification: 6,
 
     max_age_hot_post_days: 2,
 
--- a/src/backend/env/mod.rs
+++ b/src/backend/env/mod.rs
@@ -2205,9 +2205,13 @@ impl State {
         let mut active_users = 0;
         let mut bots = Vec::new();
         let mut credits = 0;
+        let mut speculative_revenue = 0;
         for user in self.users.values() {
             if user.stalwart {
                 stalwarts.push(user);
+            }
+            if user.miner {
+                speculative_revenue += user.rewards().max(0);
             }
             if now < user.last_activity + CONFIG.online_activity_minutes {
                 users_online += 1;
@@ -2262,7 +2266,7 @@ impl State {
             posts,
             comments: Post::count(self) - posts,
             credits,
-            burned_credits: self.burned_cycles,
+            burned_credits: self.burned_cycles + speculative_revenue,
             total_revenue_shared: self.total_revenue_shared,
             total_rewards_shared: self.total_rewards_shared,
             account: invoices::main_account().to_string(),
--- a/src/backend/env/mod.rs
+++ b/src/backend/env/mod.rs
@@ -1058,7 +1058,7 @@ impl State {
     pub fn minting_ratio(&self) -> u64 {
         let circulating_supply: Token = self.balances.values().sum();
         let factor = (circulating_supply as f64 / CONFIG.maximum_supply as f64 * 10.0) as u64;
-        1 << factor
+        (1 << factor) * CONFIG.difficulty_amplification
     }
 
     pub fn tokens_to_mint(&self) -> BTreeMap<UserId, Token> {
--- a/src/backend/env/mod.rs
+++ b/src/backend/env/mod.rs
@@ -204,8 +204,12 @@ pub struct State {
     #[serde(skip)]
     pub backup_exists: bool,
 
+    // TODO: delete
     #[serde(skip)]
     pub last_archive: u64,
+
+    #[serde(default)]
+    pub minting_power: HashMap<Principal, Token>,
 }
 
 #[derive(Default, Deserialize, Serialize)]
@@ -2122,6 +2126,12 @@ impl State {
             .remove(&old_principal)
             .ok_or("no principal found")?;
         self.principals.insert(new_principal, user_id);
+        let minting_power = self
+            .minting_power
+            .get(&old_principal)
+            .copied()
+            .unwrap_or_default();
+        self.minting_power.insert(new_principal, minting_power);
         let user = self.users.get_mut(&user_id).expect("no user found");
         assert_eq!(user.principal, old_principal);
         user.principal = new_principal;
--- a/src/backend/env/mod.rs
+++ b/src/backend/env/mod.rs
@@ -2874,7 +2874,6 @@ pub(crate) mod tests {
         token::mint(state, account(principal), amount);
         state.minting_mode = false;
         if let Some(user) = state.principal_to_user_mut(principal) {
-            user.balance = amount;
             user.change_rewards((amount / token::base()) as i64, "");
         }
     }
--- a/src/backend/env/mod.rs
+++ b/src/backend/env/mod.rs
@@ -550,7 +550,6 @@ impl State {
     pub fn realms_posts(
         &self,
         caller: Principal,
-        page: usize,
         offset: PostId,
     ) -> Box<dyn Iterator<Item = &'_ Post> + '_> {
         let realm_ids = match self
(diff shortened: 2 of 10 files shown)


src/backend/env/token.rs
```diff
@@ -306,12 +307,22 @@ pub fn transfer(
...
     if let Some(user) = state.principal_to_user_mut(principal) {
         if user.principal == principal {
             user.balance = balance
         } else if user.cold_wallet == Some(principal) {
             user.cold_balance = balance
+        } else {
+            unreachable!("unidentifiable principal")
+        }
+        if let Some(tokens) = minted_tokens {
+            user.minted += tokens;
```
ulan: This seems inconsistent with the existing balance accounting, which is done per principal.

Would it make sense to track `minted` also per principal? Otherwise, if the user changes the principal, the total minted for the user wouldn't match the total minted for the principal.

src/backend/env/mod.rs
```diff
@@ -1061,7 +1046,7 @@ impl State {
     pub fn minting_ratio(&self) -> u64 {
         let circulating_supply: Token = self.balances.values().sum();
         let factor = (circulating_supply as f64 / CONFIG.maximum_supply as f64 * 10.0) as u64;
-        1 << factor
+        (1 << factor) * CONFIG.difficulty_amplification
```
ulan, line 1053: Wdyt about rendering this on the token page as the minting ratio?

For example:
- Instead of `24 : 1`
- Show `(4 * 6) : 1`

src/backend/env/config.rs
```diff
@@ -187,6 +192,11 @@ pub const CONFIG: &Config = &Config {
 
     active_user_share_for_minting_promille: 10,
 
+    #[cfg(any(test, feature = "dev"))]
+    difficulty_amplification: 1,
+    #[cfg(not(any(test, feature = "dev")))]
+    difficulty_amplification: 6,
```
ulan, line 198: How did you arrive at 6 for 11 years? Also the poll results by voting power are 15 years.

src/backend/env/mod.rs
taggrx: I would not invest time into this as I consider this only a workaround and already started a new discussion for a proper solution: https://6qfxa-ryaaa-aaaai-qbhsq-cai.icp0.io/#/post/849195 Btw it would be great if you would share your thoughts as well.

src/backend/env/config.rs
taggrx: See the extended commit message: https://github.com/TaggrNetwork/Taggr/pull/28/commits/29b667a42fcffbd66af0720ef7a72d15baeb017c

I applied the same logic as on reward proposals: computed the average weighted by VP. Imo this is a more correct approach for numeric polls.

src/backend/env/token.rs
taggrx: Oh yes, I forgot to mention that changing a principal makes the minting power "forfeit". 

I really struggle to motivate myself to account for that case and add complexity:

- The modulation on yesterday's weekly data has shown already that the effects are rather weak: the amount of tokens went from ~8.6k to ~7k), when we account for principal changes it can only get weaker (if at all). 
- Due to the cap of `300` this becomes even less of a problem from the long term perspective.

WDYT?

src/backend/env/config.rs
taggrx: I think I need to clarify more: I computed that we should target about 1.3-1.5k tokens per week _at the current_ activity level, then took the last distribution numbers `9125` tokens and divided by the target number, which gives us the required amplification of 6-7. 

Please note that this is all hand-waving and only a temporary solution to curb the minting. We should change this to a proper model as soon as we agree on one.

chmllr:
LGTM

chmllr (approved):

src/backend/env/token.rs
ulan: I was thinking about it more from the design point of  view. Since a balance is just a sum of transactions of a principal, and minted is the sum of mint transactions (so a subset of transactions), it seems cleaner to have the notion of minted balance on the principal as well. 

Tracking it on the user also works, but kind of breaks the symmetry and becomes more ad-hoc IMO.

My opinion is not strong, so this approach is also okay.

src/backend/env/mod.rs
ulan: Sounds good.

src/backend/env/config.rs
ulan: I see, thanks! I was wondering if we should incrementally increase the difficulty to avoid abrupt changes? Wdyt about starting with difficulty 2 and increasing it gradually over several weeks?

src/backend/updates.rs
```diff
@@ -79,22 +79,19 @@ fn post_upgrade() {
...
-
-        // Init all users as minters
-        for user in state.users.values_mut() {
-            user.miner = true;
+        for (owner, amount) in state
+            .ledger
+            .iter()
+            .filter(|tx| tx.from.owner == Principal::anonymous())
+            .map(|tx| (tx.to.owner, tx.amount))
+            .collect::<Vec<_>>()
```
ulan, line 89: Do we really need to allocate the vector here?

ulan (approved):
Lgtm, modulo the question about increasing the difficulty more gradually.

src/backend/env/config.rs
taggrx: What would be a good reason for a gradual increase? I consider this as quick fix of a situation spiralling out of control. E.g. if we hit 5% of total supply, a [circuit-breaker](https://github.com/TaggrNetwork/Taggr/blob/main/src/backend/env/mod.rs#L1125) will trigger and Taggr will not mint anything at all. Last week we hit 3.3%.

And most importantly, as with the other suggestion, I would not like to put more complexity into this than absolutely necessary. I think this change is much better than stopping the minting as many would have supported as well. 

Given that we most likely will reduce the weekly numbers even more (and not increase anymore), there is no good reasons to invest time in adding a gradual slow down and keep minting tokens, knowing we're not happy about how they get distributed. But please let me know if you feel strongly about it.

src/backend/env/token.rs
taggrx: > it seems cleaner to have the notion of minted balance on the principal as well.

How would we use it though? For example, after this release, new principal changes will not destroy the "minting power" anymore as this field is tracked on the user level. So we're talking right now only about the previously minted minting power, which won't be accounted, but it would not matter as soon as the user reaches the donation cap.

mechaquan: Btw, what happens to the 300 cap?!  Does this go to 50 (300 / 6) because of the difficulty factor?!

taggrx: Note I simulated this change on the data from yesterday and Taggr would go to about 1.5k tokens. By Friday we would probably be at 1.8k.

src/backend/updates.rs
taggrx: Yes, we need to move all data from `state` to the heap in order to mutate `state` below.

Also note, this is a one-off code.

src/backend/env/config.rs
ulan: > And most importantly, as with the other suggestion, I would not like to put more complexity into this than absolutely necessary.

I didn't mean to do it programmatically. Instead, we could increase the factor with every release. The motivation would be to soften the impact and also observe if the posting behavior will change.

src/backend/env/token.rs
ulan: > How would we use it though? For example, after this release, new principal changes will not destroy the "minting power" anymore as this field is tracked on the user level.

I guess I am trying to say that tracking on the principal level seems more consistent. For example, everyone would be able to verify that their `minted` balance is exactly the sum of balances in their transaction history regardless whether they changed their principal or not. If we track on the user side, then the `minted` balance and their transaction history will diverge after principal changes. So purely from the design point of view, the principal tracking seems more "right".

As I said, we can go ahead with the existing solution since it works as well.

src/backend/env/config.rs
radu-daniel: And if the posting behavior changes we revert? 😀

ulan: > And if the posting behavior changes we revert?

There will be more time for discussion. I always prefer smaller steps and more iterations. With abrupt changes it is difficult to predict what can go wrong.

taggrx: Actually I thought more about this and am very confident now that it will not have any prolonged negative effects. 

First of all, I’m convinced that the current distribution is wrong and would like to enforce more merit based distributions. For example, as Ulan pointed out somewhere, I will create quarterly reward proposals for all contributions of different people and would encourage others do the same. 

Content only should not create whales in governance, IMO because we all agree in the meantime that quality content is a very relative concept and difficult to measure. 

Then, economically, by reducing the supply, at the equally staying demand, the price will just multiply by the difficult increase over time. 

And finally, those who feel (until the price increase) that it’s not worth their time to use Taggr, they IMO should leave anyway. 

With that I feel very confident doing this change because I’m convinced that it fixes things. Then the question if a quick fix is better than a slow one is kind of obvious to me.

taggrx: Oh and: we already crossed the previous Fridays number of tokens to be minted and were have 2 days to mint in front of us!


Also it looks like the proposed time based distribution I proposed will be approved without modifications. But that one is very easy to implement so we can have it very soon, sooner than phasing out the current difficulty by artificial halvings.

src/backend/env/token.rs
taggrx: I thought more about this and I think I share your point now. It’s indeed more cleaner.  Will change.

PR #30, opened 2024-03-09
feat: Implement profile pictures

ulan:
This allows a user to upload a profile picture via settings.
The profile picture is rendered in posts and comments as thumbnail.

Changes in the backend:
- add an `avatar` field to `User`.
- add an `avatar` field to `Meta`.
- add an `upload_avatar` update endpoint.
- extend existing queries to return avatar.

Changes in the frontend:
- move image helper functions to `image.ts`.
- add a profile picture field in the settings.
- show the cost of saving the settings on the button label.
- add a new `AvatarElement` helper.
- extend `UserData` to include the avatar URL.
- extend `UserLink` and `UserList` to pass user data.
- replace the user name in the header page with a new PROFILE link.
- display the profile picture in posts, comments, and profile.

### DIFF
--- a/src/backend/env/user.rs
+++ b/src/backend/env/user.rs
@@ -67,6 +67,21 @@ impl UserFilter {
     }
 }
 
+/// Contains the location of the Avatar image blob.
+#[derive(Clone, Serialize, Deserialize)]
+pub struct Avatar {
+    // The canister id of the bucket canister.
+    pub bucket_id: String,
+    pub offset: u64,
+    pub len: usize,
+}
+
+#[derive(Clone, Serialize, Deserialize)]
+pub struct UserData {
+    pub name: String,
+    pub avatar: Option<Avatar>,
+}
+
 #[derive(Serialize, Deserialize)]
 pub struct User {
     pub id: UserId,
@@ -111,6 +126,8 @@ pub struct User {
     pub miner: bool,
     #[serde(default)]
     pub controlled_realms: HashSet<RealmId>,
+    #[serde(default)]
+    pub avatar: Option<Avatar>,
 }
 
 impl User {
@@ -185,6 +202,7 @@ impl User {
             miner: true,
             downvotes: Default::default(),
             show_posts_in_realms: true,
+            avatar: None,
         }
     }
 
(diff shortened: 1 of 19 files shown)


ulan:
@taggrx  I will rebase this once your PR is merged.

src/backend/env/user.rs
```diff
@@ -67,6 +67,21 @@ impl UserFilter {
     }
 }
 
+/// Contains the location of the Avatar image blob.
+#[derive(Clone, Serialize, Deserialize)]
+pub struct Avatar {
+    // The canister id of the bucket canister.
+    pub bucket_id: String,
```
taggrx, line 74: Ouch, ouch, ouch 🙈 😂  Do you know that the current buckets implementation a pretty stupid "append only" storage and that if posts are deleted, the space is not freed up? (same would hold for avatars)

Did you think about instead simply storing these blob in the current memory, by creating an ObjectManager for these small blobs? There, both storage and removal is already supported.

src/frontend/src/settings.tsx
```diff
@@ -70,6 +85,46 @@ export const Settings = ({ invite }: { invite?: string }) => {
...
 
+    const onAvatarChange = async (ev: any) => {
+        ev.preventDefault();
+        if (!ev.target.files[0]) {
+            return;
+        }
+        const max_blob_size_bytes =
+            window.backendCache.config.max_blob_size_bytes;
+        let blob = null;
+        for (let size of [768, 512, 384, 256, 128, 64, 32]) {
```
taggrx, line 96: Is this for a better quality when downscaling?

xqxpx (commented):
Thanks Ulan, this is super cool! 🚀 I did a super high-level review, it looks great, but let's first quickly discuss the storage question. I'm curios of you considered storing in the stable memory.

src/backend/env/user.rs
ulan: Conceptually, on the high-level, I don't see difference between changing an avatar and posting a picture as long as the costs are paid. I don't think users would mind paying the photo fee.

If you'd like to optimize it, let's move avatars to the main storage. I like the effect that we can have a stable URL for avatars. I'll implement it after rebasing.

src/frontend/src/settings.tsx
ulan: Yeah, and also for simplicity. I was not in the mood for the binary search :D

ulan:
Closing in favor of the suggested alternative storage.

PR #32, opened 2024-03-12
feat: credits mode

xqxpx:
# "Normie" mode

This is the implementation of the "normie mode" for non-crypto folks as discussed and preliminary adopted in https://6qfxa-ryaaa-aaaai-qbhsq-cai.ic0.app/#/post/955992.

## Idea

Any user who gets invited, starts in the so called credits mode: any rewards they get for their posts, will only top up their credits balance which theoretically might make them use Taggr forever without top ups (if they produce good content).

Taggr does not record donations for such users and this is why they can switch to mining or rewards modes at any time.

## Limitations

Switching back to the credits mode will be only possible for users with 0 rewards (e.g. after a distribution and before getting the first reward). 

 Note that there is still a way to "cheat". E.g. by being in minting/ICP mode in the beginning of the week, then getting downvoted to 0 rewards. Then the user would still have pending donations yet they could switch to the "credits mode" due to having 0 rewards. They could then profit from credits top ups until the distribution and before the distribution switch back to ICP / mining mode. Since it only enables this cheating in a very unattractive situation (first getting downvoted to 0 rewards), I am not sure if we should protect against these edge cases on cost of complexity. But I am open to suggestions.


src/backend/env/mod.rs
```diff
@@ -1223,7 +1223,7 @@ impl State {
                 continue;
             };
             // All miner rewards are burned.
-            if user.miner {
+            if user.mode == Mode::Mining {
```
ulan: In case of `Mode::Credits` we expect the rewards to be 0, right. If that doesn't hold for some reason (e.g. a bug) would it make sense to burn those rewards?

ulan (approved):
Lgtm

src/backend/env/mod.rs
taggrx: Definitely, thanks Ulan!

meynarth:
A good one

src/backend/env/user.rs
```diff
@@ -489,7 +506,7 @@ impl User {
         principals: Vec<String>,
         filter: UserFilter,
         governance: bool,
-        miner: bool,
+        mode: Mode,
```
aligatorr89, line 509: We should do the same on user creation. Currently "Credits" mode is assigned after user creation when frontend sends `update_user`

aligatorr89: https://github.com/TaggrNetwork/Taggr/commit/aada35055e8114c9157b707b75f1307eb5322edd#r139913008

PR #34, opened 2024-03-15
Release

xqxpx:
Minor improvements and bugfixes


ulan (approved):
Lgtm

xqxpx:
@ulan I actually remembered why there existed a poll voting restriction on realm posts: because if you white list access to a realm, you probably do not want others to mess with the content of your poll. So I'll revert this change and relax to white lists restrictions only.

PR #35, opened 2024-03-16
feat: Implement profile pictures (2nd try)

ulan:
This allows a user to upload a profile picture via settings.
The profile picture is rendered in posts and comments as thumbnail.

Changes in the backend:

- add an upload_avatar update endpoint.
- store avatars in the main canister using the memory allocator.
- extend the `http_request` query to `/avatar/<user-id>`.

Changes in the frontend:

- move image helper functions to image.ts.
- add a profile picture field in the settings.
- show the cost of saving the settings on the button label.
- add a new AvatarElement helper.
- replace the user name in the header page with a new PROFILE link.
- display the profile picture in posts, comments, and profile.


ulan:
Open issues to discuss:

1. Since the URL of the avatar doesn't change anymore, the browser continue to show the cached old avatar after the user changes the avatar. It currently requires page reload to show the new image. It also doesn't allow to set long caching time for avatars. One option here would be to keep `avatar_version` field in the user object and set the url to `/avatar/<user-id>/<version>`.
2. For users that did not set their avatars, this currently shows the default avatar. It would be better to not show any image at all.

ulan:
Another open question: the images can be several KBs in size. Is it okay to store that on the main canister?

src/backend/updates.rs
```diff
@@ -235,6 +235,33 @@ fn update_user() {
...
+        if blob.len() > CONFIG.max_avatar_size_bytes {
+            return Err("avatar image is too large".to_string());
+        }
+        if blob.is_empty() {
+            if state.memory.avatars.get(&user_id).is_some() {
+                state.memory.avatars.remove(&user_id)?;
+            }
+            Ok(())
+        } else {
+            state.charge_in_realm(
```
taggrx, line 251: But why would we charge at all now? It's not a data that is stored permanently, right? Also we do not charge for the "About" section and other settings, so from the UX perspective this would be quite upsetting imo.

For example name change is different, because it is permanent and is a resource allocation (name is taken forever) so we need to prevent cybersquatting. But the avatar is just a small volatile memory.

src/backend/updates.rs
```diff
@@ -235,6 +235,33 @@ fn update_user() {
...
 
+#[update]
+async fn upload_avatar(blob: Blob) -> Result<(), String> {
+    mutate(|state| -> Result<(), String> {
+        let user_id = state.principal_to_user(caller()).ok_or("no user found")?.id;
+        if blob.len() > CONFIG.max_avatar_size_bytes {
+            return Err("avatar image is too large".to_string());
+        }
+        if blob.is_empty() {
+            if state.memory.avatars.get(&user_id).is_some() {
```
taggrx, line 246: Do we even need to check? Removal of a non-existing id is a no-op.

src/backend/updates.rs
```diff
@@ -235,6 +235,33 @@ fn update_user() {
...
+        } else {
+            state.charge_in_realm(
+                user_id,
+                CONFIG.avatar_cost,
+                None,
+                "uploading profile picture".to_string(),
+            )?;
+            if state.memory.avatars.get(&user_id).is_some() {
+                state.memory.avatars.remove(&user_id)?;
+            }
```
taggrx, line 259: Same here: does it need to be conditional? Also why not moving both of these parts below line 244 as we need to delete the previous value in any case (btw, this was a memory leak which I just fixed [here](https://github.com/TaggrNetwork/Taggr/pull/36/commits/6db50b87310300d8b25dd5cb75def5bdb1e8b9d3)).

src/frontend/src/header.tsx
```diff
@@ -194,21 +198,10 @@ export const Header = ({
     );
 };
 
-const UserSection = ({ user }: { user: UserType }) => {
+const UserSection = ({ user, hide }: { user: UserType; hide: () => void }) => {
     return (
         <div className="bottom_spaced stands_out">
             <div className="column_container centered">
-                {user && (
```
taggrx, line 201: Is it accidental?

src/frontend/src/header.tsx
```diff
@@ -219,6 +212,14 @@ const UserSection = ({ user }: { user: UserType }) => {
                             >
                                 <Journal /> JOURNAL
                             </a>
+                            <a
```
taggrx, line 215: Did you check this on mobile? I'm afraid this would be too many tiles for mobiles screen now. There was a reason, why the name was above 😅 Also because it is one of the most often clicked links for users checking their rewards (or was in the past).

src/frontend/src/header.tsx
```diff
@@ -219,6 +212,14 @@ const UserSection = ({ user }: { user: UserType }) => {
...
+                                title="PROFILE"
+                                className="icon_link"
+                                href={`/#/user/${user.id}`}
+                                onClick={hide}
+                            >
+                                <User /> PROFILE
```
taggrx, line 221: Shouldn't we have the avatar here? Actually I expected in the header which is removed now.

src/frontend/src/profile.tsx
```diff
@@ -78,7 +79,13 @@ export const Profile = ({ handle }: { handle: string }) => {
     return (
         <>
             <HeadBar
-                title={profile.name}
+                title=<>
+                    <img
+                        className="avatar_large left_well_spaced right_spaced"
+                        src={getAvatarUrl(profile.id)}
+                    ></img>
```
taggrx, line 86:
```suggestion
                    />
```

xqxpx (commented):
Thank you so much Ulan! I posted the first set of comments and questions.

src/backend/updates.rs
ulan: Uploading the image still requires some work from the canister. If users don't pay for it, then it is going to drain the cycles.

src/backend/updates.rs
ulan: You're right, a simple remove should be sufficient.

src/frontend/src/header.tsx
ulan: No, this is intentional. I thought it would make sense to keep the registration as simple as possible. New users can register with just a user name. If they want to upload an avatar, then they can do it later on in settings.

src/frontend/src/header.tsx
ulan: Good point. I haven't check this on mobile. We can keep the current layout.

src/frontend/src/header.tsx
ulan: Ok

ulan:
Thanks for review @taggrx. Looks like trying to merge a large change at once is not realistic given my time constraints. In the meantime there are more conflicts to be resolved.

I am going to split this into smaller PRs that can be merged faster.

xqxpx:
I can help with the conflicts if you think this would speed up things.

PR #36, opened 2024-03-19
Release

xqxpx:
- remove user descrimination based on the rewards balance - now a negative rewards balance only matters for minting eligibility and receiving of ICP rewards: in both cases the user needs to "pay off" the rewards debt if they get into the red. But they are not grayed out anymore
- feat: add post id into tipping memos
- feat: do not display the text field for posts on gated realms if the user is not whitelisted
- feat: show the number of unique downvoters in the last 7 days on the profile (this is more helpful for users themselves because they might have no access to realms and don't know why)
- feat: allow opening the post by clicking anywhere on the post bar
- fix: fix wrong cutting on long posts (cutting the content in the middle)

### DIFF
--- a/src/backend/env/mod.rs
+++ b/src/backend/env/mod.rs
@@ -762,7 +762,7 @@ impl State {
                 to: account(author.principal),
                 fee: Some(1), // special tipping fee
                 amount: amount as u128,
-                memo: None,
+                memo: Some(format!("Tips on post {}", post_id).as_bytes().to_vec()),
                 created_at_time: None,
             },
         )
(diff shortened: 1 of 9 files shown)


xqxpx:
Note: the memory leak bug was not effective yet as we never insert posts without removing them first. But I think for the object manager to be as close to a map as possible, reinsertion of the same id, should deallocate previously allocated memory by default.

src/backend/env/mod.rs
```diff
@@ -762,7 +762,7 @@ impl State {
                 to: account(author.principal),
                 fee: Some(1), // special tipping fee
                 amount: amount as u128,
-                memo: None,
+                memo: Some(format!("Tips on post {}", post_id).as_bytes().to_vec()),
```
chmllr, line 765: nice!

chmllr (approved):

PR #37, opened 2024-03-23
Refactoring of proposal creations

xqxpx:
This change untangles post and proposal creations.

The current implementation that evolved over time (read it is duct-taped), has the limitation that when a stalwart creates a proposal, it can only submit a text which is then used by Taggr to create a post and then attach a corresponding proposal. Unfortunately, this does not allow to attach pictures to proposals (e.g. screenshots) and stands in the way of the next feature I am working on (community roadmap).

In the refactoring we separate the proposal and post creation into different calls, refactor certain proposal types and improve the UI.

### DIFF
--- a/src/backend/env/proposals.rs
+++ b/src/backend/env/proposals.rs
@@ -101,7 +101,12 @@ impl Proposal {
                 }
             }
             Payload::Funding(_, _) => {}
-            Payload::Rewards(Rewards { votes, .. }) => {
+            Payload::Rewards(Rewards {
+                votes, receiver, ..
+            }) => {
+                if receiver == &principal {
+                    return Err("reward receivers can not vote".into());
+                }
                 let minting_ratio = state.minting_ratio();
                 let base = token::base();
                 let max_funding_amount = CONFIG.max_funding_amount / minting_ratio / base;
@@ -422,7 +427,7 @@ pub(super) fn execute_proposal(
 }
 
 #[cfg(test)]
-mod tests {
+pub mod tests {
 
     use super::*;
     use crate::{
@@ -432,6 +437,17 @@ mod tests {
         },
         STATE,
     };
+
+    pub fn propose(
+        state: &mut State,
+        caller: Principal,
+        description: String,
+        payload: Payload,
+        time: u64,
+    ) -> Result<u32, String> {
+        let post_id = Post::create(state, description, &[], caller, time, None, None, None)?;
+        create_proposal(state, caller, post_id, payload, time)
+    }
 
     #[test]
     fn test_proposal_canceling() {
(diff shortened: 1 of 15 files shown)


src/backend/env/proposals.rs
```diff
@@ -32,22 +33,36 @@ pub struct Release {
...
+#[derive(Clone, Deserialize, Serialize)]
+pub struct Rewards {
+    pub receiver: Principal,
+    pub votes: Vec<(Token, ProposedReward)>,
+    pub minted: Token,
+}
+
+// TODO: remove Clone
+#[derive(Clone, Deserialize, Serialize)]
 pub struct Reward {
```
taggrx, line 46: This struct will be removed in the next release

src/backend/env/proposals.rs
```diff
@@ -85,17 +100,8 @@ impl Proposal {
...
-                if Principal::from_text(receiver) == Ok(principal) {
-                    return Err("funding receivers can not vote".into());
-                }
-            }
-            Payload::Reward(Reward {
-                receiver, votes, ..
-            }) => {
-                if Principal::from_text(receiver) == Ok(principal) {
-                    return Err("reward receivers can not vote".into());
-                }
```
aodl: I'm still reading through things, but at the moment I'm unclear on why these safeguards have been removed (the same safeguards should apply to `Payload::Funding` and `Payload::Rewards` shouldn't they?

meynarth (approved):

src/backend/env/proposals.rs
taggrx: Not, removed, but moved to the new struct, see line 108.

Good point, I'll add the check for funding as well.

aodl: [Line 108 in the same file](https://github.com/TaggrNetwork/Taggr/pull/37/files/e66efe53d5af74f85774fecc5682616d2533515d#diff-734e075e3c202adf1c7664a452b15352864a6947f6d021a86b055f759ebc46c9L108)? That's a different validation isn't it? ("reward amount is higher than the configured maximum of {} tokens")

aodl: Ah, got ya, I was looking at an outdated version of the file (instead of [line 108](https://github.com/TaggrNetwork/Taggr/blob/d9e5475768015d5bac38b5501b59badebcaec85c/src/backend/env/proposals.rs#L108))

src/backend/env/proposals.rs
```diff
@@ -283,23 +289,31 @@ impl Payload {
...
-    description: String,
+    post_id: PostId,
     mut payload: Payload,
     time: u64,
 ) -> Result<u32, String> {
+    if Post::get(state, &post_id).is_none() {
+        return Err("post not found".to_string());
+    }
     payload.validate(state)?;
+
```
aligatorr89, line 307: Would it be better to put line 314 `!user.stalwart {` before heavy operations (like hashing the binary) ?

```suggestion
    if !user.stalwart {
        return Err("only stalwarts can create proposals".to_string());
    }
```

PR #40, opened 2024-03-29
Minor changes and bugfixes

xqxpx:


src/backend/env/user.rs
```diff
@@ -554,11 +554,10 @@ impl User {
...
     pub fn eligible_for_minting(&self) -> bool {
-        self.mode == Mode::Mining && !self.controversial() &&
-            // While `controversial` covers users with negative rewards, user with 0 rewards should
-            // not be picked for minting because they either have no rewards, or they were
-            // downvoted back to zero and then nothing should be minted for them.
-            self.rewards() > 0
+        self.mode == Mode::Mining
+            && !self.controversial()
+            && self.rewards() > 0
+            && self.active_weeks > 0
```
ulan: The white paper doesn't mention that minting is limited only to active users AFAICS. One could argue that if a user made a great post that people upvote even if the user is not active, then it might make sense to reward the user.

I am personally okay with this change, but since it changes tokenomics, would it make sense to discuss and vote first?

ulan (approved):
Looks good modulo the active users minting which I think is better to discuss first.

src/backend/env/user.rs
taggrx: Yep, I assumed no one would be bothered by this and my main worry would be that if e.g. a user has died and would never log in again, then minting tokens to their account would be kind of stupid.

But if you think this could be sensitive, I'll run a poll in /DAO, thanks!

taggrx: https://6qfxa-ryaaa-aaaai-qbhsq-cai.icp0.io/#/post/1115929

src/backend/updates.rs
```diff
@@ -81,34 +81,29 @@ fn post_upgrade() {
 #[allow(clippy::all)]
 fn sync_post_upgrade_fixtures() {
     mutate(|state| {
-        for p in state.proposals.iter_mut() {
-            p.payload = migrate_payload(p.payload.clone());
+        let controllers = state
```
taggrx, line 84: This is part of the realm editing bug fix.

src/backend/env/mod.rs
```diff
@@ -1719,33 +1718,34 @@ impl State {
...
-                if let Err(err) = self.charge(id, costs, "inactivity penalty".to_string()) {
+                if let Err(err) = self.charge(user_id, costs, "inactivity penalty".to_string()) {
                     self.logger
                         .warn(format!("Couldn't charge inactivity penalty: {:?}", err));
                 } else {
                     credits_total += costs;
                 }
             }
-            let user = self.users.get_mut(&id).expect("no user found");
-            user.change_rewards(
```
taggrx, line 1745: Not needed anymore, because who're inactive longer than for `voting_power_activity_weeks` will be in credits mode and cannot have any rewards.

PR #44, opened 2024-04-05
fix: prevent usage of uninitialized allocator

xqxpx:
This MR introduces an initialization logic to the allocator, so that we never run into the corruption issue again.

Additionally we restore all corrupted data, except for the supporter list of features. but users can simply vote on features again.

### DIFF
--- a/src/backend/updates.rs
+++ b/src/backend/updates.rs
@@ -80,39 +80,42 @@ fn post_upgrade() {
 
 #[allow(clippy::all)]
 fn sync_post_upgrade_fixtures() {
-    mutate(|state| {
-        // Restore corrupted posts 2 & 3
-        use serde_json::json;
-        let post_2 = json!({"id":2,"body":"#hello everyone! I'm [X](#/user/0) the anonymous creator of #Taggr. I wanted to build Taggr for at least 15 years but there was no point to build it on #Web2 to compete with established social platforms. After I've learned about #SNS and #InternetComputer I realized that this is exactly the right moment and the right platform: a public #blockchain on a \"sovereign physical layer\". My main goal is to make #Taggr truly decentralized and make it operate fully autonomously. Then, as a community (if there will be one) we can figure out a sustainable development path for #Taggr.","user":0,"timestamp":1640952386200311995_u64,"children":[151],"parent":null,"watchers":[],"tags":["InternetComputer","SNS","Taggr","Web2","blockchain","hello"],"reactions":{"10":[18446744073709551614_u64,18446744073709551615_u64],"100":[18446744073709551615_u64]},"patches":[],"files":{},"tree_size":1,"tree_update":0,"report":null,"tips":[],"extension":null,"realm":null,"hashes":[],"reposts":[],"heat":0});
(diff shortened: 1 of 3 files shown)


src/backend/updates.rs
```diff
@@ -79,7 +79,41 @@ fn post_upgrade() {
...
+fn sync_post_upgrade_fixtures() {
+    mutate(|state| {
+        // Restore corrupted posts 2 & 3
+        use serde_json::json;
+        let post_2 = json!({"id":2,"body":"#hello everyone! I'm [X](#/user/0) the anonymous creator of #Taggr. I wanted to build Taggr for at least 15 years but there was no point to build it on #Web2 to compete with established social platforms. After I've learned about #SNS and #InternetComputer I realized that this is exactly the right moment and the right platform: a public #blockchain on a \"sovereign physical layer\". My main goal is to make #Taggr truly decentralized and make it operate fully autonomously. Then, as a community (if there will be one) we can figure out a sustainable development path for #Taggr.","user":0,"timestamp":1640952386200311995_u64,"children":[151],"parent":null,"watchers":[],"tags":["InternetComputer","SNS","Taggr","Web2","blockchain","hello"],"reactions":{"10":[18446744073709551614_u64,18446744073709551615_u64],"100":[18446744073709551615_u64]},"patches":[],"files":{},"tree_size":1,"tree_update":0,"report":null,"tips":[],"extension":null,"realm":null,"hashes":[],"reposts":[],"heat":0});
```
aligatorr89: Anyway to verify content ? Posts and journal both fail 

```
window.api.query("journal", "X", 100, 0).then(console.log)
window.api.query("posts", [2]).then(console.log);
```

src/backend/updates.rs
```diff
@@ -79,7 +79,41 @@ fn post_upgrade() {
...
+            .posts
+            .reinsert(2, serde_json::from_value(post_2).unwrap());
+        state
+            .memory
+            .posts
+            .reinsert(3, serde_json::from_value(post_3).unwrap());
+    });
+
+    // Restore features
+    mutate(|state| state.memory.features.reset());
```
aligatorr89: we will loose `supporters` state. Shouldn't it be copied and populated ?

aligatorr89 (changes requested):

src/backend/updates.rs
aligatorr89: it changed my local posts :(

taggrx: Yes, they fail because these posts are corrupted, that's why we're restoring them. But @ulan verified with his local backup already. I hope he can confirm here.

src/backend/updates.rs
taggrx: Yes, we lost the supporters. There is no way to restore. But that's not a big deal, as it is just one button click for users.

src/backend/updates.rs
aodl: good point. Looks like this would also overwrite [post 2](https://e4i5g-biaaa-aaaao-ai7ja-cai.icp0.io/#/post/2) and [post 3](https://e4i5g-biaaa-aaaao-ai7ja-cai.icp0.io/#/post/3) in staging.

Can anyone other than X with a backup confirm that these hardcoded posts are consistent with the originals?

Presumably these posts became corrupt as part of the Roadmap memory corruption, or something else?

Am I right in understanding that the stable-structures library recently introduced is supposed to protect against stable-structure sharing memory? I'm not sure I understand how the posts were corrupted.

aligatorr89: Only ones that have previouselly downloaded backup. I think it's still acceptable if we can avoid running it on staging. 

Yes, also TLDR how it happened would be nice :D

aligatorr89:
@taggrx , found this small vulnerability. It's probably worth it to fix it here https://taggr.link/post/1194634

src/backend/updates.rs
taggrx: How it happened: currently, when we deserialize the state, we also need to initialize the stable memory allocator for all ObjectManager's so that they have a _common_ view on which regions of memory are free and which are not. 

I forgot to add this for the new object manager for features, so it was acting like the whole memory is free.

The current change introduces a panic on any allocator that was not explicitly initialized.

PR #45, opened 2024-04-05
Release

xqxpx:
Minor bugfixes

### DIFF
--- a/src/backend/env/config.rs
+++ b/src/backend/env/config.rs
@@ -346,7 +346,7 @@ pub const CONFIG: &Config = &Config {
         (12, 1),
     ],
 
-    max_funding_amount: 1_000_000, // at ratio 1:1
+    max_funding_amount: 2_000_000, // at ratio 1:1
 
     neuron_id: 16737374299031693047,
 };
--- a/src/backend/updates.rs
+++ b/src/backend/updates.rs
@@ -82,12 +82,9 @@ fn post_upgrade() {
 fn sync_post_upgrade_fixtures() {
     mutate(|state| {
         // Compensate for the executed by failed proposal #/post/1159871
-        state
-            .realms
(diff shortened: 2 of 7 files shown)


src/backend/updates.rs
```diff
@@ -79,7 +79,17 @@ fn post_upgrade() {
 }
 
 #[allow(clippy::all)]
-fn sync_post_upgrade_fixtures() {}
+fn sync_post_upgrade_fixtures() {
+    mutate(|state| {
+        // Compensate for the executed by failed proposal #/post/1159871
+        state
```
aligatorr89: deploying this locally resulted in error, had to comment it out, probably same will happen on staging

aligatorr89 (changes requested):

src/backend/updates.rs
taggrx: Very good point, thanks! Will make it fail-safe.

src/backend/env/config.rs
```diff
@@ -346,7 +346,7 @@ pub const CONFIG: &Config = &Config {
         (12, 1),
     ],
 
-    max_funding_amount: 1_000_000, // at ratio 1:1
+    max_funding_amount: 2_000_000, // at ratio 1:1
```
aodl, line 349: Am I right in understanding that Fund proposals are not capped, and that this cap applies just to Reward proposals?

I wonder if it could do with being renamed to `max_reward_amount`, or the cap applied to fund proposals as well (if this is the intention).

aligatorr89 (approved):

aodl (approved):
LGTM

src/backend/env/config.rs
taggrx: No, it is the same cap for both.

PR #48, opened 2024-04-17
Pre release

xqxpx:


radu-daniel (commented):
Should/could we add a character counter?

xqxpx:
> Should/could we add a character counter?

Why? Are you referencing the increased costs? In this case, Taggr will show the correct cost estimation to you.

src/backend/env/mod.rs
```diff
@@ -274,6 +274,42 @@ pub enum Destination {
 }
 
 impl State {
+    pub fn delay_weekly_chores(&mut self, caller: Principal) -> bool {
```
leo-ship-it, line 277: Should we delay the chores by one week only? Why not one day or any other delay?

src/backend/env/mod.rs
```diff
@@ -274,6 +274,42 @@ pub enum Destination {
...
+            return false;
+        }
+
+        self.weekly_chores_delay_votes.insert(user.id);
+
+        if self.weekly_chores_delay_votes.len() * 100
+            / self.users.values().filter(|user| user.stalwart).count()
+            >= CONFIG.report_confirmation_percentage as usize
+        {
+            self.last_weekly_chores += WEEK;
```
leo-ship-it, line 296: Shouldn't we reset `self.weekly_chores_delay_votes`?

Also, I think with the current writing a stalwart could delay the weekly chores indefinitely as soon as the condition in L.292 is verified just by voting yes again.

src/backend/env/mod.rs
```diff
@@ -1624,33 +1660,20 @@ impl State {
...
-            mutate(|state| {
-                state
-                    .logger
-                    .info("Skipping minting & distributions due to open proposals")
-            });
-        }
 
+            state.minting_mode = true;
+            state.mint();
+            state.minting_mode = false;
```
leo-ship-it, line 1666: Why do we set the minting mode to true then false?

src/backend/env/mod.rs
```diff
@@ -4676,17 +4682,47 @@ pub(crate) mod tests {
...
+            for i in 0..9 {
+                assert_eq!(state.last_weekly_chores, 0);
+                assert!(state.delay_weekly_chores(pr(i)));
+            }
+
+            // shifting happened
+            assert_eq!(state.last_weekly_chores, WEEK);
+
+            // more votes are rejected
+            assert!(!state.delay_weekly_chores(pr(10)));
```
leo-ship-it, line 4718: Check if the shift can happen again if a stalwart triggers it again.

leo-ship-it:
General question, why did you delay the minting if some proposal was open? Because IIUC, the minting happens synchronously, or maybe you want to avoid interrupting the ICP distribution.

src/backend/env/mod.rs
taggrx: >Shouldn't we reset self.weekly_chores_delay_votes?

As answered above, since I assumed we will always need an upgrade and the field is marked as `skip`, an upgrade will automatically reset it. We can definitely reset the field on the next (delayed minting), but I am not sure if this is a good and realistic scenario. As I said, this operation is consider as a total out-of-order situation.

>Also, I think with the current writing a stalwart could delay the weekly chores indefinitely as soon as the condition in L.292 is verified just by voting yes again.

I don't follow. It's a hashmap, a repeated voting will be a noop. Am I missing something?

src/backend/env/mod.rs
taggrx: My assumption was that we would only ever use this mechanism if we determine some defect or vulnerability in Taggr's smart-contract. This would always require an upgrade. Since we cannot predict how long a fix would take, one week sounds like a safe bet.

src/backend/env/mod.rs
taggrx: The minting logic checks deeps inside for this flag being set. It's just one countermeasure to avoid triggering minting (which is nothing else but a transfer from an anonymous principal) from unexpected places.

xqxpx:
> General question, why did you delay the minting if some proposal was open? Because IIUC, the minting happens synchronously, or maybe you want to avoid interrupting the ICP distribution.

Because an open proposal has a count down which counts 1% less per day from the total voting power in order to get any proposal approved within 99 days. IF we would mint, we would increase the total VP again without any possibility to resubmit a vote with the increased VP. This could create unpredictable effects. But since the current minting is very low and gets even lower soon, this is not a concern anymore.

src/backend/env/mod.rs
taggrx: added

src/backend/env/mod.rs
leo-ship-it: Ok my bad, actually if we already shifted then we reject any new votes. Everything looks good, my comment about the test is not relevant.

src/backend/updates.rs
```diff
@@ -80,80 +80,18 @@ fn post_upgrade() {
...
-
-    // Compensate for the executed by failed proposal #/post/1159871
-    mutate(|state| {
-        if let Some(realm) = state.realms.get_mut("RUGANG") {
-            realm.controllers.insert(1277);
+            let _ = Post::mutate(state, &1180586, |post| {
+                post.report.take();
+                Ok(())
+            });
+            let Some(user) = state.users.get_mut(&3750) else {
```
ulan: Did you mean 3570? https://6qfxa-ryaaa-aaaai-qbhsq-cai.ic0.app/#/user/3570
3750 is CristiNU, not KeydaKorven.

Nit: using `if let Some(user) {...}` seems more readable to me compared to `return`.

src/backend/updates.rs
```diff
@@ -80,80 +80,18 @@ fn post_upgrade() {
...
+                return;
+            };
+            let _ = user.change_credits(1000, user::CreditsDelta::Plus, "report cancellation");
+            state.burned_cycles -= 1000;
```
ulan, line 94: Is it okay if it goes below zero?

src/backend/env/mod.rs
```diff
@@ -274,6 +274,42 @@ pub enum Destination {
...
+        else {
+            return false;
+        };
+
+        // If we shifted already, exit.
+        if self.last_weekly_chores >= time() + WEEK {
+            return false;
+        }
+
+        self.weekly_chores_delay_votes.insert(user.id);
```
aligatorr89, line 290: even if it's HashSet we should still throw error if user already voted

aligatorr89 (changes requested):

src/backend/env/mod.rs
aligatorr89: Sounds good, each deploy will reset `weekly_chores_delay_votes.` We can also do it with own post upgrade function if needed.

src/backend/env/mod.rs
aligatorr89: since it's an extreme sitation, it's acceptable as it is IMO

src/backend/updates.rs
taggrx: Yes, negative balance is actually possible: this means the debt of the DAO. We even had this in the past, where charging of canisters drained more cycles than the revenue.

src/backend/updates.rs
taggrx: Oh wow, thanks a lot Ulan! Hm, I was under the impression that I _copied_ the user id from the url...  😅 

Fixed!

aligatorr89 (approved):

aligatorr89 (approved):

PR #49, opened 2024-04-22
Release: reaction cost reductions, feature closing functionality and bugfixes

xqxpx:
Most important changes of this release:

- Implements a feature closing functionality for features on the roadmap (a release can now contain a list of feature ids that get closed with the corresponding code change).
- Decreases costs as [decided](https://taggr.link/#/post/1246146) by the DAO.
- Disables proposal creation of NNS proposal posts for network economics and replica upgrades.
- Bug fixes.


src/backend/env/post.rs
```diff
@@ -769,10 +769,9 @@ pub fn archive_cold_posts(state: &mut State, max_posts_in_heap: usize) -> Result
     state.backup_exists = false;
 
     state.logger.debug(format!(
-        "`{}` posts archived (posts still in heap: `{}`, instructions used: `{}B`)",
+        "`{}` posts archived (posts still in heap: `{}`)",
```
aodl, line 772: Why remove the extra info from this log?

docs/WHITEPAPER.md
```diff
@@ -22,9 +22,9 @@ Below is a breakdown of costs.
...
 | On-chain pictures            |   `B * $blob_cost` | For `B` pictures in a post or comment                                                                      |
 | Poll                         |       `$poll_cost` | For adding a poll to a post or comment                                                                     |
-| Reacting with ❤️ , 👍, 😢    |                `3` | Gives `2` reward points, burns the rest as a fee.                                                          |
-| Reacting with 🔥, 😂, 🚀, 💯 |               `12` | Gives `10` rewards points, burns the rest as a fee.                                                        |
-| Reacting with ⭐️, 🏴‍☠️        |               `23` | Gives `20` reward points, burns the rest as a fee.                                                         |
+| Reacting with ❤️ , 👍, 😢    |                `2` | Gives `1` reward points, burns the rest as a fee.                                                          |
+| Reacting with 🔥, 😂, 🚀, 💯 |                `5` | Gives `5` rewards points, burns the rest as a fee.                                                         |
```
aodl: Is this meant to be '6' credits instead of '5', otherwise there's not fee to burn. Is that the intention?

src/frontend/src/realms.tsx
```diff
@@ -266,7 +266,7 @@ export const RealmForm = ({ existingName }: { existingName?: string }) => {
                             timer = setTimeout(async () => {
                                 // @ts-ignore
                                 realm.controllers = await userNameToIds(
-                                    input.split(","),
+                                    input.split(/[,\s]/),
                                 );
                                 setRealm({ ...realm });
                             }, 1500);
```
aodl, line 272: I wonder if this timer should also be increased to 3 seconds instead of 1.5. If you're using a mobile device then typing may not be as slick an experience (if you don't type the name is quickly then it'll disappear before you have a chance to finish it)

src/backend/env/mod.rs
```diff
@@ -1517,7 +1517,7 @@ impl State {
             .filter(|proposal| proposal.id > last_known_proposal_id)
         {
             // Vote only on proposals with topics network economics, governance, SNS & replica-management.
```
aodl: Does this comment need adjusting to account for the change in relevant proposal topics to avoid future confusion?

src/backend/env/post.rs
taggrx: Because now the outer log contains this number for the entire routine.

docs/WHITEPAPER.md
taggrx: Nice catch, thank you!

src/frontend/src/realms.tsx
taggrx: You think 1.5s is too low to hit the next letter on mobile? Are you sure?

src/backend/env/mod.rs
taggrx: Absolutely, thanks a lot!

src/frontend/src/realms.tsx
aodl: I had a go last night and if you're typing in a long name or one you need to stop and think about half way through, then it's tricky (but if no-one else experiences this when using there mobile, then I guess its not of much concern)

src/backend/env/mod.rs
aodl: Any chance of reverting this change for now though? I think an automated due diligence notice and a link to CodeGov reviews could improve matters here. I'm happy to take a look at doing this

src/frontend/src/realms.tsx
taggrx: Ideally we should have a dropdown of course...

src/backend/env/mod.rs
taggrx: Oh you're saying you want to actually vote on these proposal according to CodeGov's vote?

aodl: Not necessarily, but providing a reference to their reviews for others (on TAGGR) to take a look at would help users make an more informed decision. I can get my ICPPuppy account to automatically post a comment on each proposal for this.

I'd plan on reading what the CodeGov reviewers have to say, take a look myself to see if there's anything worth commenting on, verifying the build, and then voting myself. At the very least, having a bunch of people that verify the build before voting seems useful. With time and more users we can gradually make this process more robust.

src/frontend/src/realms.tsx
aodl: Yeah I guess so, although could be cumbersome as the list of users grows, unless it's a search/dropdown box. I miss the feature that got removed recently where recognised names are displayed beneath the input field (then the unrecognised names could get purged from the input field on submission, or when the field loses focus, rather than on a timer). Anyway, no major issue, just seemed worth commenting on.

src/backend/env/mod.rs
taggrx: Ok, let me remove the change for now.

src/backend/updates.rs
```diff
@@ -335,6 +325,7 @@ fn propose_release(post_id: PostId, commit: String, binary: ByteBuf) -> Result<u
                 commit,
                 binary: binary.to_vec(),
                 hash: Default::default(),
+                closed_features: features,
```
aligatorr89, line 328: Validation if post id includes feature. I can provide any random posts here.

PR #51, opened 2024-05-13
Release

xqxpx:
- Bugfix: do not show post 0 as a closing feature in a release
- Unify canister top up and use ICP for all canisters.

### DIFF
--- a/src/backend/updates.rs
+++ b/src/backend/updates.rs
@@ -79,18 +79,11 @@ fn post_upgrade() {
 }
 
 #[allow(clippy::all)]
-fn sync_post_upgrade_fixtures() {
-    mutate(|state| {
-        // Reset Keyda's report again (the last time missed `post_report`, see https://taggr.link/#/post/1273612)
-        if let Some(user) = state.users.get_mut(&3570) {
-            user.post_reports.clear();
-            user.report.take();
-        }
-    })
-}
+fn sync_post_upgrade_fixtures() {}
 
 #[allow(clippy::all)]
 async fn async_post_upgrade_fixtures() {
+    // We need to upgrade the buckets in this release.
     storage::upgrade_buckets().await;
 }
 
--- a/src/bucket/src/lib.rs
+++ b/src/bucket/src/lib.rs
@@ -1,16 +1,9 @@
 use candid::{CandidType, Deserialize, Principal};
-use ic_cdk::{
-    api::{
-        self,
-        call::{arg_data_raw, reply_raw},
-        canister_balance,
-        management_canister::{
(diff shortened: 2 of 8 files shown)


src/bucket/src/lib.rs
```diff
@@ -68,27 +61,6 @@ fn balance() -> u64 {
...
-        cycles.0.to_u64_digits().last().copied().unwrap_or_default(),
-        idle_cycles_burned_per_day
-            .0
-            .to_u64_digits()
-            .last()
-            .copied()
-            .unwrap_or(1),
-    )
-}
-
```
leo-ship-it, line 91: I think it can still be useful to access the cycles of the canister, for ckBTC I added this endpoint to fetch the canister status: https://github.com/dfinity/ic/blob/master/rs/bitcoin/ckbtc/minter/src/main.rs#L190

leo-ship-it (approved):
LGTM

src/bucket/src/lib.rs
taggrx: For that we have the `balance` endpoint :) (https://github.com/TaggrNetwork/Taggr/pull/51/files/8b4a5e983c8a48e677f6cc69501db727e3ad8592#diff-aa9d115efb38645a0536dc70df4be252f6d47a4d92283cee2d66ddf133c51fe3L67)

chmllr:
One problem we have now is that we need to watch two canisters instead of just one, if we want to top them up with cheap cycles, right? 

Otherwise LGTM.

aodl:
LGTM

aligatorr89 (approved):

src/backend/updates.rs
```diff
@@ -79,18 +79,11 @@ fn post_upgrade() {
...
-            user.post_reports.clear();
-            user.report.take();
-        }
-    })
-}
+fn sync_post_upgrade_fixtures() {}
 
 #[allow(clippy::all)]
 async fn async_post_upgrade_fixtures() {
+    // We need to upgrade the buckets in this release.
```
aligatorr89, line 86: On this note, is this why i'm getting tons of these (since last release):

image

xqxpx:
> One problem we have now is that we need to watch two canisters instead of just one, if we want to top them up with cheap cycles, right?
> 
> Otherwise LGTM.

That's a good catch, thanks Christian! I changed it to only top up the children canisters with 1 day of cycles budget. Since it's happening on an hourly basis, this should be good enough.

src/backend/updates.rs
taggrx: If you didn't upgrade the buckets - yes.

PR #52, opened 2024-05-17
Release

xqxpx:
Minor bug fixes: see commit descriptions

### DIFF
--- a/src/backend/dev_helpers.rs
+++ b/src/backend/dev_helpers.rs
@@ -62,9 +62,28 @@ async fn clear_buckets() {
 #[update]
 fn replace_user_principal(principal: String, user_id: UserId) {
     mutate(|state| {
-        state
-            .principals
-            .insert(Principal::from_text(principal).unwrap(), user_id)
+        use crate::token::Account;
+        let principal = Principal::from_text(principal).unwrap();
+        state.principals.insert(principal, user_id);
+        let user_principal = state.principal_to_user(principal).unwrap().principal;
+        let balance = state
+            .balances
+            .remove(&Account {
+                owner: user_principal,
+                subaccount: None,
+            })
+            .unwrap();
+        state.balances.insert(
+            Account {
+                owner: principal,
+                subaccount: None,
+            },
+            balance,
+        );
+        let power = state.minting_power.remove(&user_principal).unwrap();
+        state.minting_power.insert(principal, power);
+        let user = state.principal_to_user_mut(principal).unwrap();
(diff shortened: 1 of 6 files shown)


src/backend/dev_helpers.rs
```diff
@@ -62,9 +62,28 @@ async fn clear_buckets() {
 #[update]
 fn replace_user_principal(principal: String, user_id: UserId) {
```
aligatorr89, line 63: I don't see this method being exposed anywhere?

aligatorr89 (approved):

src/backend/env/mod.rs
```diff
@@ -271,6 +271,17 @@ pub enum Destination {
 }
 
 impl State {
+    pub fn register_post_tags(&mut self, post_id: PostId, tags: &BTreeSet<String>) {
+        for tag in tags {
```
aligatorr89, line 275: It's missing removing tags. If i remove tag from post, it still shows it when searching for it

aligatorr89 (changes requested):

src/backend/dev_helpers.rs
chmllr: I think those are helper methods — see the file name.

chmllr (approved):
LGTM

aodl (commented):
Why move post archiving to the daily routine? It looks like the hourly routine tends to take more than a few minutes. Presumably this change is intended to bring that down to less than a minute to reduce noise in the hourly logs? Would an alternative approach be to increase the threshold for hourly logs slightly?

xqxpx:
@aodl right it is to decrease the verbosity. Also it takes less than 0B instructions, so there is no danger in moving it to daily for now. I used these logs to watch the stability, but see no issues anymore, so this procedure is simply not really interesting.

src/backend/env/mod.rs
taggrx: That's fine, the deleted posts are not rendered in most feed anyway. also I think it's a good to see that the post existed. However, also note that tag indexes are rotating. They only keep the last 1k posts. So any de;eted post will disappear eventually anyway, this is why I see no reasons to complicate the code.

aodl: But what about posts that are not deleted, but are edited to remove erroneously added tags? I think this is what @aligatorr89 was describing

aligatorr89: @taggrx , I described removing tags from post (edit post)

PR #53, opened 2024-05-30
feat: token auctions

xqxpx:
This PR introduces a weekly auction. We start with only 15 TAGGR. The tokens get minted to the auction winner if there is a sell out. The established market price is printed to the logs on the dashboard and is not used yet for anything.

### DIFF
--- a/src/backend/env/auction.rs
+++ b/src/backend/env/auction.rs
@@ -1,0 +1,393 @@
+use std::collections::BTreeSet;
+
+use candid::Principal;
+use ic_ledger_types::{
+    AccountIdentifier, Memo, Subaccount, Tokens, DEFAULT_FEE, DEFAULT_SUBACCOUNT,
+};
+use serde::{Deserialize, Serialize};
+
+use crate::{
+    env::invoices::{self},
+    id, mutate, read, time,
+};
+
+use super::{
+    invoices::{main_account, principal_to_subaccount},
+    token::Token,
+    user::UserId,
+    State, Time,
+};
+
+pub const AUCTION_ICP_SUBACCOUNT: [u8; 32] = [
+    0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1, 0, 1,
+];
+
+pub fn auction_account() -> AccountIdentifier {
+    AccountIdentifier::new(&id(), &Subaccount(AUCTION_ICP_SUBACCOUNT))
+}
+
+#[derive(Clone, Eq, Debug, Default, PartialEq, Serialize, Deserialize)]
+pub struct Bid {
+    pub user: UserId,
+    pub amount: Token,
+    // e8s for 1 TAGGR cent
+    pub e8s_per_token: u64,
+    timestamp: Time,
+}
+
+impl PartialOrd for Bid {
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for Bid {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        if self.e8s_per_token != other.e8s_per_token {
+            // prioritize higher bids
+            return self.e8s_per_token.cmp(&other.e8s_per_token);
+        }
+        if self.timestamp != other.timestamp {
+            // prioritize older bids
+            return other.timestamp.cmp(&self.timestamp);
+        }
+        if self.amount != other.amount {
+            // prioritize larger bids
+            return self.amount.cmp(&other.amount);
+        }
+        self.user.cmp(&other.user)
+    }
+}
+
+#[derive(Default, Serialize, Deserialize)]
+pub struct Auction {
+    pub amount: Token,
+    bids: BTreeSet<Bid>,
+}
+
+impl Auction {
+    /// Checks if there is a sell out.
+    pub fn sell_out(&self) -> bool {
+        self.bids.iter().map(|bid| bid.amount).sum::<u64>() >= self.amount
+    }
+
+    /// Returns the highest bids if they form a sell out.
+    pub fn get_bids(&mut self) -> Vec<Bid> {
+        if !self.sell_out() {
+            return Default::default();
+        }
+        let mut amount = self.amount;
+        let mut bids = Vec::default();
+
+        while amount > 0 {
+            let Some(bid) = self.bids.pop_last() else {
+                break;
+            };
+            let eff_amount = bid.amount.min(amount);
+            bids.push(Bid {
+                user: bid.user,
+                amount: eff_amount,
+                e8s_per_token: bid.e8s_per_token,
+                timestamp: bid.timestamp,
+            });
+            // if the bid was larger than tokens left, create a leftover bid
+            if eff_amount < bid.amount {
+                self.bids.insert(Bid {
+                    user: bid.user,
+                    amount: bid.amount - eff_amount,
+                    e8s_per_token: bid.e8s_per_token,
+                    timestamp: bid.timestamp,
+                });
+            }
+            amount = amount.saturating_sub(eff_amount);
+        }
+
+        bids
+    }
+}
+
+/// Cancels user's bid and returns the funds to user wallet.
+pub async fn cancel_bid(principal: Principal) -> Result<u64, String> {
+    let bid = mutate(|state| remove_bid(state, principal))?;
+
+    let user_account = read(|state| {
+        let user_principal = state
+            .principal_to_user(principal)
+            .expect("no user found")
+            .principal;
+        AccountIdentifier::new(&user_principal, &DEFAULT_SUBACCOUNT)
+    });
+
+    let funds = bid
+        .amount
+        .checked_mul(bid.e8s_per_token)
+        .expect("overflow")
+        .checked_sub(DEFAULT_FEE.e8s())
+        .expect("nothing to refund");
+
+    invoices::transfer(
+        user_account,
+        Tokens::from_e8s(funds),
+        Memo(727),
+        Some(Subaccount(AUCTION_ICP_SUBACCOUNT)),
+    )
+    .await
+    .map_err(|err| {
+        let msg = format!("couldn't withdraw funds from bid {:?}: {}", bid, err);
+        mutate(|state| state.logger.error(&msg));
+        msg
+    })
+}
+
+fn remove_bid(state: &mut State, principal: Principal) -> Result<Bid, String> {
+    let user_id = state
+        .principal_to_user(principal)
+        .ok_or("no user found")?
+        .id;
+
+    let bid = state
+        .auction
+        .bids
+        .iter()
+        .find(|bid| bid.user == user_id)
+        .ok_or("no bids found")?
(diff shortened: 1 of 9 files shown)


leo-ship-it:
Another approach would be to fetch `quote` from the Taggr/ICP pool on ICPSwap. It would be easier and wouldn't require any extra front-end work. What do you think?
https://dashboard.internetcomputer.org/canister/opl73-raaaa-aaaag-qcunq-cai

src/backend/env/auction.rs
```diff
@@ -0,0 +1,393 @@
...
+    }
+    let user_id = state
+        .principal_to_user(principal)
+        .ok_or("no user found")?
+        .id;
+
+    assert!(
+        !state.auction.bids.iter().any(|bid| bid.user == user_id),
+        "no bids exist for the user"
+    );
```
chmllr, line 210: Maybe use `has_bid` here?

src/backend/env/auction.rs
```diff
@@ -0,0 +1,393 @@
...
+pub async fn create_bid(
+    principal: Principal,
+    amount: Token,
+    e8s_per_token: u64,
+) -> Result<(), String> {
+    // cancel existing bid if necessary
+    if read(|state| has_bid(state, principal)) {
+        cancel_bid(principal).await?;
+    }
+
```
chmllr, line 177: Maybe we should `assert!(!has_bid(state, principal));` here and after the transfer to avoid any kind of potential races?

chmllr (approved):
LGTM

src/backend/env/config.rs
```diff
@@ -163,6 +165,8 @@ pub const CONFIG: &Config = &Config {
     logo: include_str!("../../frontend/assets/logo.min.svg"),
     staging: "e4i5g-biaaa-aaaao-ai7ja-cai.icp0.io",
 
+    weekly_auction_size_tokens: 15000,
+
```
aodl, line 169: I thought the idea was for only 15 TAGGR to be auctioned for now. 15000 / `tokenBase` == 150 (not 15).

Is this intentionally using different units?

src/backend/env/auction.rs
```diff
@@ -0,0 +1,393 @@
...
+        .amount
+        .checked_mul(bid.e8s_per_token)
+        .expect("overflow")
+        .checked_sub(DEFAULT_FEE.e8s())
+        .expect("nothing to refund");
+
+    invoices::transfer(
+        user_account,
+        Tokens::from_e8s(funds),
+        Memo(727),
```
aodl, line 131: Is this 727 meaningful? or the 717 and 737 used as memos for the other transactions. Is it worth defining these in one place as named constants (along with any other memo currently used elsewhere - 999, 0x50555054, 828282)?

src/backend/env/auction.rs
```diff
@@ -0,0 +1,393 @@
...
+    /// Returns the highest bids if they form a sell out.
+    pub fn get_bids(&mut self) -> Vec<Bid> {
+        if !self.sell_out() {
+            return Default::default();
+        }
+        let mut amount = self.amount;
+        let mut bids = Vec::default();
+
+        while amount > 0 {
+            let Some(bid) = self.bids.pop_last() else {
```
aodl, line 83: It looks like leftover bids are not cleared for the next auction (which seems intentional given the re-insertion of a remainder bid below).

If I've understood this correctly, as long as there is at least one *disinterested* but rational user on TAGGR with some minuscule amount of ICP that they're willing to part with, it makes sense for them to submit a bid for the maximum amount of TAGGR for the minimum possible ICP price.

If there's ever a scenario where the auction would otherwise be cancelled due to not enough bids, what would actually happen is that this described user would receive all of the auctioned TAGGR for a minuscule amount of ICP - screwing up the established market price (rather than the auction being cancelled due to lack of competition). Have I understood this correctly?

src/backend/env/config.rs
taggrx: You are right, thanks! Hm, this is strange as I definitely had 15 TAGGR in the local testing and I don't remember changing it. I will check this part again.

src/backend/env/auction.rs
taggrx: They have no meaning right now, I just use different numbers to be able to distinguish txs in the future if needed (ever). Then yes we could have a docs or something like that.

src/backend/env/auction.rs
taggrx: Yes, the bids are kept for the next auction. The behaviour you describe is possible if no interest in minted tokens exist from other users. But in this case, the described trick is still possible by anybody to do, it's just a little bit easier.

At the same time, if this happens, no one cares about the token and it is the real market price, right?

Note that the minting from the price will still be capped, so it's just the user who had this bid will get a lot of TAGGR for free (and it would be deserved), but nothing catastrophic should happen.

WDYT?

src/backend/env/auction.rs
taggrx: I see no good reason for this. This behaviour can only be triggered by a broken client and the second assert will still not prevent the "loss" of funds, right?

src/backend/env/auction.rs
aodl: > if this happens, no one cares about the token

Maybe, but the surest way to maximise the amount of TAGGR minted would be for everyone to agree to bid the lowest amount of ICP for the highest amount of TAGGR. You would not necessarily loose out even if your bid was not the first to be submitted. What would be more important is that you *purchase* super cheap TAGGR afterwards (before the cap is reached). I could imagine peer pressure and transparency causing the community to potentially cooperate in this way.

The users that stand to loose out are those that already own lots of TAGGR (inflation due to maximal minting). This would motivate/require such users to place more reasonable bids even though they don't necessarily want any more TAGGR. Due to bid transparency, I could see this leading to more of a 'them and us' mentality (causing division between new and established users - the have and the have nots)

aodl: I feel like it would be good to have a safety measure, such as querying the [Exchange Rate Canister](https://dashboard.internetcomputer.org/canister/uf6dk-hyaaa-aaaaq-qaaaq-cai) (or another approach for establishing the external market rate for TAGGR) and make sure that the internal market rate (from auctioning) is never less than some arbitrary percentage of the external market rate (say... 90%). This allows for anywhere between 0% and 10% saving by purchasing from the TAGGR platform directly.

This still delivers on generating revenue for the platform, while allowing users to benefit from somewhat cheaper TAGGR, while also protecting against significant price manipulation due to some sort of abuse of the auction.

taggrx: > would be for everyone to agree to bid the lowest amount of ICP for the highest amount of TAGGR

I personally plan to set a weekly timer to check for arbitrage opportunities. Additionally, I'll set a catch-order on BEACON for TAGGR say 25% below the market price. This order can sit forever there. Once filled, again: arbitrage. This alone will make the arbitraging very easy for everyone. I think I won't be alone doing this: this is literally free money, after all. So I can't really imagine how the attack you described can ever work... It assumes, people don't need free money 🤷‍♂️

> is never less than some arbitrary percentage of the external market rate 

But having an "external" price ratio kills the whole purpose of the auction - being autonomous, because now we need to fetch it from somewhere. There should be no knowledge of external market rate: there are simply no good enough decentralized external markets right now. If there would be, we wouldn't need the auction.

aodl: My feeling was that the point of the auction was to establish a *fair* price that would tend to be slightly lower than the going market rate, providing a reason for users to purchase their TAGGR from the platform (which mints it) rather than from external exchanges (thereby generating more revenue for TAGGR).

I like the idea of avoiding external dependencies, but it just feels to me like getting users to internally establish the market rate that TAGGR will be minted by incentivises those users to try and make it as cheap as possible, so that they can purchase more of it more cheaply. As far as I can tell, the only thing stopping this from happening would be a *tragedy of the commons* type of situation (which is usually referenced as a bad thing), whereby two users notice that all the bids are offering extremely low amounts of ICP for high amounts of TAGGR, so one of them decides to swoop in and bid for all the auctioned TAGGR at a more reasonable rate, but then the other one does the same (say this happens a few times, so the bids become more reasonable). But now because of those two users' actions, the established market rate that TAGGR can be minted by has shot up. Sure, one of them got 15 TAGGR minted to them as a result of the auction, but if they'd just played the cooperative game that everyone else was trying to play, the internal market rate for TAGGR would have been crazy slow, allowing them to purchase significantly more TAGGR much more cheaply after the auction closed.

Given that the established market rate isn't used for selling TAGGR with this release (other than the auction), it wouldn't currently be an issue anyway. But it's something I can imagine being a problem whenever that release comes out. If feels like users stand to gain more by cooperatively gaming the auction, than by competing in the auction - simply because after the auction has closed then would be able to buy TAGGR based on the results of the auction

taggrx: Couldn't we actually fix it as following: after the minting price is established, every user can decide to convert the collected rewards to TAGGR at this price or to ICP? This way, no matter what happens to the auction price, users have the last word wrt. to their own minting.

But yes, I would address this in the next step or so.

src/backend/env/config.rs
taggrx: Found the discrepancy! I initially used the value of 150, but in all local tests 15 (just a local change), but forgot to change globally. But I think we should start with 15, not 150. So I pushed an update.

src/backend/env/auction.rs
aodl: > every user can decide to convert the collected rewards to TAGGR at this price or to ICP

But if the auction has been exploited cooperatively to maximise the amount that can be minted to everyone cheaply (after the auction, by purchasing from the platform), everyone would rightly choose TAGGR instead of ICP.

I think this problem is inherent in using an auction for a small amount of TAGGR to establish the price for a large amount of TAGGR. Why are we wanting to allow for the purchasing of TAGGR after the auction anyway (I think this is the problem)? Why not make the auction the only means of aquiring TAGGR from the platform? This should increase interest in the auction, and therefore make it more competitive.

What do you think?

I also think it would be possible to detect how competitive the auction is at the current amount of auctioned TAGGR (e.g. 15) and increase it or decrease it automatically on the next auction based on whether or not the previous auction was competitive enough. Looking at the number of bidders wouldn't be very sybil resistant, but you could take the average TAGGR rate of successful bids, then check if the cheapest successful bid is within some distance of the average. If it deviates significantly, it indicates low competition and the amount of TAGGR being auctioned could be dropped to protect the price. If the deviation is minimal, the amount of TAGGR being auctioned could potentially be increased (up to a maximum cap) to speed up governance decentralisation to some desirable target.

**Just thinking out loud**, I'm not trying to object to this specific release as it doesn't yet support purchasing of TAGGR after the auction. Given that this is an experiment at this stage I think this release is a positive step in a good direction 👍

taggrx: > Why not make the auction the only means of aquiring TAGGR from the platform? This should increase interest in the auction, and therefore make it more competitive.
> What do you think?

Hm, this is an interesting idea, actually. We need to discuss this on Taggr, with everyone!

> increase it or decrease it automatically on the next auction based on whether or not the previous auction was competitive enough

This is another good idea! However I still think it's not possible that there will be no sell out, because everyone can instantly dump the newly minted Taggr on a market. So it doesn't matter if people collude or not, it take one arbitrageur to do this trade.

Anyway, these discussions are very interesting, so let's merge this PR and continue on Taggr with the new observations?

PR #55, opened 2024-06-16
Release: new market price based minting

xqxpx:


src/backend/env/mod.rs
```diff
@@ -1160,90 +1162,57 @@ impl State {
...
             title: "Token minting report".into(),
             description: Default::default(),
             items: Vec::default(),
         };
 
+        let miner_rewards = self
+            .users
+            .values_mut()
+            .filter(|user| user.mode == Mode::Mining)
+            .fold(HashMap::new(), |mut acc, user| {
```
ulan: I don't see why we need a `HashMap` here. Looks like we iterate the users and visit each user only once.

src/backend/env/mod.rs
```diff
@@ -1160,90 +1162,57 @@ impl State {
...
+                    return acc;
+                }
+                acc.insert(user.id, rewards);
+                acc
+            });
+
+        // burn a corresponding amount credits to generate revenue
+        self.burned_cycles += miner_rewards.values().sum::<i64>();
+
+        let tokens_to_mint: HashMap<_, _> = miner_rewards
```
ulan: Ditto about the `HashMap`.

ulan: Could this iteration be fused into the iteration above?

src/backend/env/mod.rs
```diff
@@ -1160,90 +1162,57 @@ impl State {
...
-        tokens_to_mint
-    }
-
-    pub fn mint(&mut self) {
-        let circulating_supply: Token = self.balances.values().sum();
-        if circulating_supply >= CONFIG.maximum_supply {
+    /// Takes the market_price (e8s per 1 token) and mints new tokens for all miners with positive
+    /// rewards according to the market price ratio.
+    pub fn mint(&mut self, market_price: u64) {
+        if market_price == 0 {
```
ulan, line 1168: Would it make sense to set the reasonable lower and upper bound for sanity checks?
Should we add a log why minting is being skipped here?

aligatorr89:
Found this weird behaviour https://taggr.network/thread/1375187

src/backend/env/mod.rs
```diff
@@ -1160,93 +1162,69 @@ impl State {
...
         let total_tokens_to_mint: u64 = tokens_to_mint.values().sum();
 
-        if ratio > 1
-            && total_tokens_to_mint * 100 / circulating_supply > CONFIG.minting_threshold_percentage
-        {
+        if total_tokens_to_mint > CONFIG.max_funding_amount {
             self.logger.warn(format!(
-                "Skipping minting: `{}` tokens exceed the configured threshold of `{}`% of existing supply.",
-                total_tokens_to_mint, CONFIG.minting_threshold_percentage
+                "Skipping minting: new tokens amount `{}` exceeds the configured threshold of `{}`",
```
ulan: We have already taken rewards and burned cycles at this point. Should we revert those changes when skipping? Or better: check for the max funding before doing the changes?

ulan (commented):
Nice simplification!

src/backend/env/mod.rs
chmllr: I actually wonder if this check is correct at all. Imagine Taggr gets slashdotted and generates a lot of revenue in a week, that would correspond to more than 1k tokens. Why would we skip the minting? Based on which logic?

If we want to throttle the distribution, then maybe we should throttle the minting: i.e. as soon as we reached 1k tokens, we simply stop processing rewards and they become sort of a backlog for the next minting?

chmllr:
LGTM

src/backend/env/mod.rs
taggrx: Agreed, will collapse and simplify, thanks Ulan!

src/backend/env/mod.rs
taggrx: Will add!

src/backend/env/mod.rs
taggrx: Good point, actually! I think having an early return on reaching the threshold is indeed the safest we can do for now without impacting legitimate cases... Thanks Christian! Will change.

PR #56, opened 2024-06-21
fix: timers and preview window

xqxpx:
Groups all infos about timers in one struct: the last execution time and pending status. This will now prevent overlapping routines even in cases where routines are overlapping.

### DIFF
--- a/src/backend/env/mod.rs
+++ b/src/backend/env/mod.rs
@@ -143,9 +143,19 @@ pub struct Summary {
     items: Vec<String>,
 }
 
+#[derive(Default, Clone, Serialize, Deserialize)]
+pub struct Timers {
+    pub last_weekly: Time,
+    pub last_daily: Time,
+    pub last_hourly: Time,
+
+    pub weekly_pending: bool,
+    pub daily_pending: bool,
+    pub hourly_pending: bool,
+}
+
 #[derive(Default, Serialize, Deserialize)]
 pub struct State {
-    #[serde(default)]
     pub auction: Auction,
 
     pub burned_cycles: i64,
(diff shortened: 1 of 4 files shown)


src/backend/updates.rs
```diff
@@ -80,7 +80,12 @@ fn post_upgrade() {
 
 #[allow(clippy::all)]
 fn sync_post_upgrade_fixtures() {
-    mutate(|state| state.auction.amount = CONFIG.weekly_auction_size_tokens)
+    mutate(|state| {
+        state.timers.last_hourly = state.last_hourly_chores;
+        state.timers.last_daily = state.last_daily_chores;
+        // compensate for the bug introduced by 3 overlapping weekly routines
+        state.timers.last_weekly = state.last_weekly_chores - 2 * WEEK;
```
aligatorr89, line 87: last weekly ran 4 times, why do we need to reset time ?

aligatorr89: NVM, I see last was 5 of July :D

aligatorr89 (approved):

src/backend/env/mod.rs
```diff
@@ -1671,18 +1683,26 @@ impl State {
             }
         };
 
-        if last_weekly_chores + WEEK < now {
+        if timers.last_weekly + WEEK < now && !timers.weekly_pending {
+            mutate(|state| {
+                state.timers.weekly_pending = true;
```
ulan, line 1688: If the canister traps during execution of the chore, then it will never retry the chore again due to the pending flag. Is the plan for such cases to reset the flags with an upgrade?

ulan (approved):
Looks good thanks!

src/backend/env/mod.rs
taggrx: This is correct. I remember I've read somewhere that there is a way to do some clean ups on panics too, but I couldn't find [anything useful](https://github.com/dfinity/cdk-rs/blob/7fa27b03d0d891b30a89b1b281ca3ba0661f2750/src/ic-cdk/src/api/call.rs#L114) in the rust CDK. So, currently I prefer a manual fix via upgrade, but mostly because I do not know yet how to do better... Any advice?

ulan: There is a "guard" pattern that uses the `drop` destructor to cleanup code after panics. See for example: https://github.com/dfinity/ic/blob/9cad8170f98b8fd58f392b2395adf0a7f1d3e3c2/rs/ethereum/cketh/minter/src/guard/mod.rs

PR #60, opened 2024-07-27
Release

xqxpx:
- Fixes revenue accounting from the auction (with retroactive applying for the auction from 26th of July)
- Fixes the UI bug showing. the wrong number of max rewards in a proposal (found by @i4eky)
- Fixes VP caching in reward proposals.

### DIFF
--- a/src/backend/env/mod.rs
+++ b/src/backend/env/mod.rs
@@ -1760,7 +1760,14 @@ impl State {
             return 0;
         }
 
+        mutate(|state| state.distribute_revenue_from_icp(revenue));
+
         market_price
+    }
+
+    pub fn distribute_revenue_from_icp(&mut self, e8s: u64) {
+        self.burned_cycles +=
+            (e8s as f64 / self.e8s_for_one_xdr as f64 * CONFIG.credits_per_xdr as f64) as i64;
     }
 
     fn distribute_realm_revenue(&mut self, now: Time) {
@@ -3006,6 +3013,18 @@ pub(crate) mod tests {
         state
             .new_user(p, 0, name.to_string(), Some(credits))
             .unwrap()
+    }
+
+    #[test]
+    fn test_revenue_from_icp() {
+        mutate(|state| {
+            state.e8s_for_one_xdr = 13510000;
+            assert_eq!(state.burned_cycles, 0);
+            // distribute 69 ICP
+            state.distribute_revenue_from_icp(6907960000);
+            // collect roughly 511k credits (6907960000/13510000)
+            assert_eq!(state.burned_cycles, 511321);
+        })
     }
 
     #[test]
--- a/src/backend/env/proposals.rs
+++ b/src/backend/env/proposals.rs
@@ -106,10 +106,6 @@ impl Proposal {
                 if receiver == &principal {
                     return Err("reward receivers can not vote".into());
                 }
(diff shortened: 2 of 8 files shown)


src/backend/updates.rs
```diff
@@ -80,7 +80,12 @@ fn post_upgrade() {
 }
 
 #[allow(clippy::all)]
-fn sync_post_upgrade_fixtures() {}
+fn sync_post_upgrade_fixtures() {
+    mutate(|state| {
+        // Distribute revenue from the last auction: https://dashboard.internetcomputer.org/transaction/6b959af7ad051d15f78d818bd27b53ccc51f84e825e0ee9a940582a20860be23?index=13096034
+        state.distribute_revenue_from_icp(6907960000);
```
ulan, line 86: I would vote for not fixing it and keeping it in TAGGR mainly because I don't know how to verify the account addresses in the transaction above and more importantly I don't want to spend time doing that :smile:

src/backend/env/proposals.rs
```diff
@@ -93,11 +99,17 @@ impl Proposal {
...
             Payload::Rewards(Rewards {
-                votes, receiver, ..
+                submissions,
+                receiver,
+                ..
             }) => {
                 if receiver == &principal {
                     return Err("reward receivers can not vote".into());
                 }
+                let user_id = state
```
ulan: Nit: we have `user` available and could use `user.id`.

src/backend/updates.rs
taggrx: Sorry, I don't understand what you mean. Keeping what in TAGGR? The TAGGR tokens auctioned were minted to bidders. But now we have the ICP from the auction sitting in the Treasury that we need to distribute.

You can verify that we need to distribute 69 ICP easily using the Taggr Dashboard only (https://6qfxa-ryaaa-aaaai-qbhsq-cai.icp0.io/#/dashboard) and the fact we know we auctioned 115 TAGGR:

1. The dashboard says, the established market price is `60069300` e8s (115 * 60069300 = ~69 ICP).
2. Click on the Treasury link on the dashboard which leads you to the treasury ICP account and make sure you see the transaction linked above being executed at `2024-07-26, 19:02:55 UTC` which is the UTC distribution time. 

We of course could not restore anything, but I feel a bit like a betrayer after advertising the new revenue source in the [last proposal](https://6qfxa-ryaaa-aaaai-qbhsq-cai.icp0.io/#/post/1385856) 😅

src/backend/env/proposals.rs
taggrx: Great, thanks so much!

src/backend/updates.rs
anonymous-qubit: Thanks for the detailed explanation! Now I can follow and verify. The retrofix lgtm!

src/backend/env/mod.rs
```diff
@@ -1760,9 +1760,16 @@ impl State {
             return 0;
         }
 
+        mutate(|state| state.distribute_revenue_from_icp(revenue));
+
         market_price
     }
 
+    pub fn distribute_revenue_from_icp(&mut self, e8s: u64) {
```
aligatorr89, line 1768: function name suggests it distributes to users, but that will happen later

`account_revenue_from_icp`

PR #63, opened 2024-08-26
Release

xqxpx:
- Bugfixes
- Random reward of 10 TAGGR with the winning chance proportional to credits spent during the last week


src/backend/env/user.rs
```diff
@@ -415,6 +419,7 @@ impl User {
                     .cycles
                     .checked_sub(amount)
                     .ok_or("wrong negative delta amount")?;
+                self.credits_burned += amount;
```
ulan: Are these credits actually burned? IIUC, `change_credits` is also called when doing various operations with credits. For example, calling the `transfer_credits` endpoint seems to call this function.

If that's the case, then it is easy to game this lottery by just transferring credits between two accounts, no?

IMO, the credits burned should come from the useful user activity such as publishing a post or reacting.

taggrx: Thanks Ulan, yeah I realized yesterday night that there exists a similar issue with reports: the credits are actually burned when the user got reported, but this is not what we want to reward by increasing chances of a lottery...

So I will need to change this to only count the credits on content creation + reactions and nothing else.

taggrx: Please check the new commit. We now count only credits spent on positive reactions and new posts. We don't count excess penalties, negative reactions and nothing else where credits might be spent (like invites, etc).

taggrx: PS: only the fees (actually burned credits) are counted on reactions. The part that effectively gets transferred to the rewarded user will not count.

src/backend/env/mod.rs
```diff
@@ -1727,6 +1725,59 @@ impl State {
...
+                    !user.controversial()
+                        && user.active_within_weeks(time(), 1)
+                        && user.credits_burned() > 0
+                }) {
+                    threshold += user.take_credits_burned();
+                    allocation.push((user.id, threshold));
+                }
+
+                // Truncate the random number so that every single user has a chance to win.
+                random_number %= threshold + 1;
```
ulan: There seems to be some off-by-one issue in this distribution.

Let's take the simplest possible case of two users where each user burned 1 credit.

- threshold = 2
- random_number can take values 0, 1, 2
- the first user wins if the random number is 0 or 1
- the second user winds if the random number is 2

So the distribution is biased towards the first user.

A fix would be:
```
if threshold == 0 { bail out}
random_number %= threshold;
```

In the `find` predicate use the strict comparison: `threshold > random_number`.

ulan (approved):
Looks good modulo the comment about the random distribution.

src/backend/env/mod.rs
taggrx: Yeah, sounds good to me! Thanks Ulan!

PR #64, opened 2024-09-04
Release

xqxpx:
- Moves the token ledger to the stable memory
- Sets the transaction fees to 0.1
- Sets the tupping fees to 0
- Adds memos to all minting transactions
- Code clean up (removes password migration)


src/backend/updates.rs
```diff
@@ -56,7 +56,6 @@ fn post_upgrade() {
         }
     }
     stable_to_heap_core();
-    mutate(|state| state.load());
```
taggrx, line 59: This was a bug: we had a redundant state initialization.

src/backend/env/memory.rs
```diff
@@ -82,50 +84,68 @@ impl Memory {
...
         };
         self.api_ref = Rc::new(RefCell::new(test_api));
         self.posts.init(Rc::clone(&self.api_ref));
         self.features.init(Rc::clone(&self.api_ref));
-    }
-
-    #[cfg(test)]
-    pub fn unpack_for_testing(&mut self) {
-        self.unpack();
+        self.ledger.init(Rc::clone(&self.api_ref));
```
chmllr: Couldn't we just `self.api = test_api` and then call `self.init()` here?

src/backend/env/mod.rs
```diff
@@ -200,6 +201,9 @@ pub struct State {
     #[serde(skip)]
     pub last_upgrade: u64,
 
+    #[serde(skip)]
+    pub token_fees_burned: Token,
```
chmllr, line 205: So now we will only update it on upgrades?

chmllr (approved):

src/backend/updates.rs
```diff
@@ -83,7 +82,15 @@ fn post_upgrade() {
 fn sync_post_upgrade_fixtures() {}
 
 #[allow(clippy::all)]
-async fn async_post_upgrade_fixtures() {}
+async fn async_post_upgrade_fixtures() {
```
chmllr: It should be in `sync_post_upgrade_fixtures` right?

src/backend/env/memory.rs
taggrx: It turns out we could not because the initialization of the allocator was happening during cloning, which is a very convoluted logic, so I refactored this part and now we can. Thanks Christian!

src/backend/env/mod.rs
taggrx: Yes, I think this is good enough. This metric is not required to be real-time.

src/backend/updates.rs
taggrx: Yep, thanks!

PR #66, opened 2024-09-12
Release

xqxpx:
- Avataggrs
- ICRC3 support

### DIFF
--- a/src/backend/queries.rs
+++ b/src/backend/queries.rs
@@ -49,7 +49,7 @@ fn features() {
     read(|state| {
         let ids: Vec<PostId> = parse(&arg_data_raw());
         reply(
-            features::features(state, ids)
+            features::features(state, &ids)
                 .map(|(post_id, tokens, feature)| {
                     Post::get(state, &post_id).map(|post| (post.with_meta(state), tokens, feature))
                 })
@@ -138,7 +138,7 @@ fn transactions() {
             iter.rev()
                 .skip(page * CONFIG.feed_page_size)
                 .take(CONFIG.feed_page_size)
-                .collect::<Vec<(u32, _)>>(),
+                .collect::<Vec<(&u32, _)>>(),
         );
     });
 }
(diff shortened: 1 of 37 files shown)


src/backend/queries.rs
```diff
@@ -49,7 +49,7 @@ fn features() {
     read(|state| {
         let ids: Vec<PostId> = parse(&arg_data_raw());
         reply(
-            features::features(state, ids)
+            features::features(state, &ids)
```
aligatorr89, line 52: After this update, i'm facing this issue with `features()`

`Panicked at 'couldn't deserialize: ErrorImpl { code: EofWhileParsingValue, offset: 22 }', src/backend/env/memory.rs:65:40.`

aligatorr89 (approved):

aligatorr89:
Couple of old issues: 
- Issue with transactions from minting account show 0.25fee. This is fixed in this version. https://taggr.network/#/transaction/50931
- I am able to tip myself: https://taggr.network/#/transaction/51297: not fixed

aligatorr89:
If tip doesn't carry cost in Taggr, can it at least cost like 2 credits? Anyone can indefinetelly tip back and fourth

src/backend/queries.rs
taggrx: Could you please post the reproduction steps?

xqxpx:
> If tip doesn't carry cost in Taggr, can it at least cost like 2 credits? Anyone can indefinetelly tip back and fourth

Yes, that a good point! I actually wanted to implement exactly that, but forgot about it. Thanks so much for flagging this!

src/backend/queries.rs
aligatorr89: I deployed this version, and the issue started occurring.

taggrx: Maybe you are upgrading from a wrong version? Since we skipped the previous release,you need to upgrade from the currently deployed version - 29-08-24.

PR #67, opened 2024-10-06
Feature request: Invites revamp

aligatorr89:
Feature request: https://taggr.network/#/post/1333906

- Extend invites to struct: credits, credits_per_user, realm_id, joined_by_user_ids
- Change create_invite
   - add new parameters and validate edge cases
- Update update_invite:
  -  owner can increase credits allowed to consume on invite
  -   owner can change realm_id
- Implement frontend requirements
  - Revamp invite page
  - Registration flow: if invite has realm_id, navigate user to that realm. Avoid showing him Mint credits page
- Protect transfering credits to 1 account (any of below it passes) `user::validate_send_credits`
  - invited user consumed amount of credits from invite
  - invited user purchased new credits
- Addtionally: create_invite has small vulnerability, anyone can create Infinite number of invites without paying anything. It can consume heap memory. Fixed within `invite::validate_user_invites_credits`
  - checking if ALL invites credits are smaller than owners amount of credits

### DIFF
--- a/src/backend/env/invite.rs
+++ b/src/backend/env/invite.rs
@@ -79,7 +79,10 @@ impl Invite {
     }
 }
 
-pub fn invites_by_principal(state: &State, principal: Principal) -> Vec<(&String, &Invite)> {
+pub fn invites_by_principal(
+    state: &State,
+    principal: Principal,
+) -> Box<impl Iterator<Item = (&String, &Invite)>> {
     let invites = state
         .principal_to_user(principal)
         .ok_or("Principal not found")
@@ -88,16 +91,17 @@ pub fn invites_by_principal(state: &State, principal: Principal) -> Vec<(&String
             state
                 .invite_codes
                 .iter()
-                .filter(|(_, invite)| invite.inviter_user_id == user_id)
-                .collect()
+                .filter(move |(_, invite)| invite.inviter_user_id == user_id)
         })
         .expect("Failed to filter invites");
 
-    invites
+    Box::new(invites)
 }
 
 /**
  * Check allocated credits in invites do not exceed user's credits balance
+ *
+ * Protects against creating infinite number of invites
  */
 pub fn validate_user_invites_credits(
     state: &State,
--- a/src/backend/env/invite.rs
+++ b/src/backend/env/invite.rs
@@ -8,8 +8,7 @@ pub struct Invite {
     pub credits_per_user: Credits,
     pub joined_user_ids: BTreeSet<UserId>,
     pub realm_id: Option<RealmId>,
-    /// Owner id
-    pub user_id: UserId,
+    pub inviter_user_id: UserId,
 }
 
 impl Invite {
@@ -17,7 +16,7 @@ impl Invite {
         credits: Credits,
         credits_per_user: Credits,
         realm_id: Option<RealmId>,
-        user_id: UserId,
+        inviter_user_id: UserId,
     ) -> Self {
         // Convert empty realm_id to None
         let converted_relm_id: Option<RealmId> = realm_id.filter(|id| !id.is_empty());
@@ -27,20 +26,20 @@ impl Invite {
             credits_per_user,
             joined_user_ids: BTreeSet::new(),
             realm_id: converted_relm_id,
-            user_id,
+            inviter_user_id,
         }
     }
 
     pub fn consume(&mut self, joined_user_id: UserId) -> Result<(), String> {
+        if self.joined_user_ids.contains(&joined_user_id) {
+            return Err("User already credited".into());
+        }
+
         let new_credits = self
             .credits
             .checked_sub(self.credits_per_user)
             .ok_or("Invite credits too low")?;
         self.credits = new_credits;
-
-        if self.joined_user_ids.contains(&joined_user_id) {
-            return Err("User already credited".into());
-        }
 
         self.joined_user_ids.insert(joined_user_id);
 
@@ -53,11 +52,18 @@ impl Invite {
         realm_id: Option<RealmId>,
         user_id: UserId,
     ) -> Result<(), String> {
-        if self.user_id != user_id {
+        if self.inviter_user_id != user_id {
             return Err("Owner does not match".into());
         }
 
         if let Some(new_credits) = credits {
+            if new_credits % self.credits_per_user != 0 {
+                return Err(format!(
+                    "Credits per user are not a multiple of credits {} {}",
+                    new_credits, self.credits_per_user
+                ));
+            }
+
             self.credits = new_credits;
         }
 
@@ -80,13 +86,16 @@ pub fn invites_by_principal(state: &State, principal: Principal) -> Vec<(String,
             state
                 .invite_codes
                 .iter()
-                .filter(|(_, invite)| invite.user_id == user.id)
+                .filter(|(_, invite)| invite.inviter_user_id == user.id)
                 .map(|(code, invite)| (code.clone(), invite.clone()))
                 .collect::<Vec<_>>()
         })
         .unwrap_or_default()
 }
 
+/**
+ * Check allocated credits in invites do not exceed user's credits balance
+ */
 pub fn validate_user_invites_credits(
     state: &State,
     user: &User,
@@ -100,38 +109,22 @@ pub fn validate_user_invites_credits(
     let total_invites_credits: Credits = state
         .invite_codes
         .values()
-        .filter(|invite| invite.user_id == user.id)
+        .filter(|invite| invite.inviter_user_id == user.id)
         .map(|invite| invite.credits)
         .sum();
 
     let total_with_diff = total_invites_credits
         .checked_add(new_credits)
         .ok_or("Invite credits overflow")?
-        .checked_sub(old_credits.unwrap_or_default());
+        .checked_sub(old_credits.unwrap_or_default())
+        .ok_or("Invite credits overflow")?;
 
-    match total_with_diff {
-        Some(total_with_diff) => {
-            if total_with_diff > user.credits() {
-                return Err(format!(
-                    "You don't have enough credits to support invites {} , {}",
-                    user.credits(),
-                    total_with_diff
-                ));
-            }
-        }
-        None => {
-            return Err("Invite credits overflow".into());
-        }
-    }
-
-    Ok(())
-}
-
-pub fn validate_realm_id(state: &State, realm_id: Option<&RealmId>) -> Result<(), String> {
-    if let Some(id) = realm_id {
-        if !id.is_empty() && !state.realms.contains_key(id) {
-            return Err(format!("Realm {} not found", id.clone()));
-        };
+    if total_with_diff > user.credits() {
+        return Err(format!(
+            "You don't have enough credits to support invites {} , {}",
+            user.credits(),
+            total_with_diff
+        ));
     }
 
     Ok(())
--- a/src/backend/env/invite.rs
+++ b/src/backend/env/invite.rs
@@ -131,10 +131,42 @@ pub fn validate_user_invites_credits(
 pub(crate) mod tests {
     use crate::{
         mutate,
-        tests::{create_user, pr},
+        tests::{create_user, create_user_with_credits, pr},
+        Realm,
     };
 
     use super::*;
+
+    /// Creates invite with 200 credits, "test" realm, 50 credits per user
+    pub fn create_invite_with_realm(
+        state: &mut State,
+        principal: Principal,
+    ) -> (UserId, String, RealmId) {
+        let id = create_user_with_credits(state, principal, 2000);
+        state
+            .create_realm(
+                principal,
+                "test".into(),
+                Realm {
+                    controllers: vec![id].into_iter().collect(),
+                    ..Default::default()
+                },
+            )
+            .expect("realm creation failed");
+        state
+            .create_invite(principal, 200, Some(50), Some("test".to_string()))
+            .expect("invite creation failed");
+
+        (
+            id,
+            invites_by_principal(state, principal)
+                .last()
+                .expect("invite not found")
+                .0
+                .clone(),
+            "test".to_string(),
+        )
+    }
 
     #[test]
     fn test_invite_validation() {
@@ -174,5 +206,95 @@ pub(crate) mod tests {
                 Err("not enough credits available: 1000 (needed for invites: 1089)".into())
             );
         })
+    }
+
+    #[test]
+    fn test_update() {
+        let (user_id, code, realm_id) = mutate(|state| create_invite_with_realm(state, pr(1)));
+
+        // Update credits and realm
+        mutate(|state| {
+            let invite = state.invite_codes.get_mut(&code).expect("invite not found");
+            // Unset realm
+            assert_eq!(invite.update(None, Some("".to_string()), user_id), Ok(()));
+            assert_eq!(invite.credits, 200);
+            assert_eq!(invite.realm_id, None);
+            // Set different credits and realm
+            assert_eq!(
+                invite.update(Some(250), Some(realm_id.clone()), user_id),
+                Ok(())
+            );
+            assert_eq!(invite.credits, 250);
+            assert_eq!(invite.realm_id, Some(realm_id.clone()));
+        });
+
+        // Unset to 0 is not allowed unless it was used at least once
+        mutate(|state| {
+            let invite = state.invite_codes.get_mut(&code).expect("invite not found");
+            assert_eq!(
+                invite.update(Some(0), None, user_id),
+                Err("cannot set credits to 0 as it has never been used".into())
+            );
+
+            invite.joined_user_ids.insert(101); // Mock
+            assert_eq!(invite.update(Some(0), None, user_id), Ok(())); // Pass
+        });
+
+        // Credits per user have to be mutliple of credits
+        mutate(|state| {
+            let invite = state.invite_codes.get_mut(&code).expect("invite not found");
+            assert_eq!(
+                invite.update(Some(140), None, user_id),
+                Err("credits per user 50 are not a multiple of new credits 140".into())
+            );
+        });
+
+        // Owner does not match
+        let other_user = mutate(|state| create_user(state, pr(5)));
+        mutate(|state| {
+            let invite = state.invite_codes.get_mut(&code).expect("invite not found");
+            assert_eq!(
+                invite.update(Some(200), None, other_user),
+                Err("owner does not match".into())
+            );
+        });
+    }
+
+    #[test]
+    fn test_consume() {
+        let (_, code, invitee_id_1) = mutate(|state| {
+            let (user_id, code, _) = create_invite_with_realm(state, pr(6));
+            let invitee_id_1 = create_user(state, pr(7));
+            (user_id, code, invitee_id_1)
+        });
+
+        mutate(|state| {
+            let invite = state.invite_codes.get_mut(&code).expect("invite not found");
+            // Consume credits
+            assert_eq!(invite.consume(invitee_id_1), Ok(()));
+            assert_eq!(invite.credits, 150);
+            // Already credited
+            assert_eq!(
+                invite.consume(invitee_id_1),
+                Err("user already credited".into())
+            );
+        });
+    }
+
+    #[test]
+    #[should_panic]
+    fn test_consume_panic() {
+        let (_, code, invitee_id_1) = mutate(|state| {
+            let (user_id, code, _) = create_invite_with_realm(state, pr(8));
+            let invitee_id_1 = create_user(state, pr(9));
+            (user_id, code, invitee_id_1)
+        });
+
+        let _ = mutate(|state| {
+            let invite = state.invite_codes.get_mut(&code).expect("invite not found");
+            // Credits too low
+            invite.credits = 10;
+            invite.consume(invitee_id_1)
+        });
     }
 }
--- a/src/backend/env/invite.rs
+++ b/src/backend/env/invite.rs
@@ -64,6 +64,11 @@ impl Invite {
                 ));
             }
 
+            // Protect against creating invite and setting to 0 without usage
+            if new_credits == 0 && self.joined_user_ids.is_empty() {
+                return Err("Cannot set credits to 0 as it has never been used".into());
+            }
+
             self.credits = new_credits;
         }
 
--- a/src/backend/env/invite.rs
+++ b/src/backend/env/invite.rs
@@ -32,13 +32,13 @@ impl Invite {
 
     pub fn consume(&mut self, joined_user_id: UserId) -> Result<(), String> {
         if self.joined_user_ids.contains(&joined_user_id) {
-            return Err("User already credited".into());
+            return Err("user already credited".into());
         }
 
         let new_credits = self
             .credits
             .checked_sub(self.credits_per_user)
-            .ok_or("Invite credits too low")?;
+            .ok_or("invite credits too low")?;
         self.credits = new_credits;
 
         self.joined_user_ids.insert(joined_user_id);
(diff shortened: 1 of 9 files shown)


aligatorr89:
Had small issues because timer is mocked to 0 in tests. I had to revert some new tests due to this issue

xqxpx:
Will review soon! 👍

src/backend/env/invite.rs
```diff
@@ -0,0 +1,138 @@
...
+use std::collections::BTreeSet;
+
+#[derive(Default, Serialize, Deserialize, Clone)]
+pub struct Invite {
+    pub credits: Credits,
+    pub credits_per_user: Credits,
+    pub joined_user_ids: BTreeSet<UserId>,
+    pub realm_id: Option<RealmId>,
+    /// Owner id
+    pub user_id: UserId,
```
taggrx: Should we use a name so that 1) the comment is not needed and 2) the semantics is clear on any line where it is used, e.g. `inviter`?

src/backend/env/invite.rs
```diff
@@ -0,0 +1,138 @@
...
+
+impl Invite {
+    pub fn new(
+        credits: Credits,
+        credits_per_user: Credits,
+        realm_id: Option<RealmId>,
+        user_id: UserId,
+    ) -> Self {
+        // Convert empty realm_id to None
+        let converted_relm_id: Option<RealmId> = realm_id.filter(|id| !id.is_empty());
```
taggrx: Is this the right way, though? Shouldn't we instead check that the realm exist at all? 

If there is no easy way to do it here, alternatively, can we handle the case later? E.g. if the realm_id is present, but no realm can be found, we default to the None case?

I just see no good reason for handling one special case for `""` while it still can be `"abra-cadabra"` :)

src/backend/env/invite.rs
```diff
@@ -0,0 +1,138 @@
...
+    pub fn consume(&mut self, joined_user_id: UserId) -> Result<(), String> {
+        let new_credits = self
+            .credits
+            .checked_sub(self.credits_per_user)
+            .ok_or("Invite credits too low")?;
+        self.credits = new_credits;
+
+        if self.joined_user_ids.contains(&joined_user_id) {
+            return Err("User already credited".into());
+        }
```
taggrx: Please move this to the top of the function. If we'd exit with an error here, we leave the state mutated (credits subtracted).

src/backend/env/invite.rs
```diff
@@ -0,0 +1,138 @@
...
+            self.credits = new_credits;
+        }
+
+        if let Some(id) = realm_id {
+            if id.is_empty() {
+                self.realm_id = None;
+            } else {
+                self.realm_id = Some(id);
+            }
+        }
```
taggrx: WDYT, how real is this case? To me it feels like a 99.99% edge case, but I might be wrong. In case you agree: should we maybe simplify and not allow it at all? If somebody would need it, they can create a new invite.

Just in case you agree: I'd also rename the function to `top_up`.

src/backend/env/invite.rs
```diff
@@ -0,0 +1,138 @@
...
+            return Err("Invite credits overflow".into());
+        }
+    }
+
+    Ok(())
+}
+
+pub fn validate_realm_id(state: &State, realm_id: Option<&RealmId>) -> Result<(), String> {
+    if let Some(id) = realm_id {
+        if !id.is_empty() && !state.realms.contains_key(id) {
```
taggrx: Wouldn't `state.realms.contains_key("")` return `false` anyway?

src/backend/env/invite.rs
```diff
@@ -0,0 +1,138 @@
...
+
+pub fn invites_by_principal(state: &State, principal: Principal) -> Vec<(String, Invite)> {
+    state
+        .principal_to_user(principal)
+        .map(|user| {
+            state
+                .invite_codes
+                .iter()
+                .filter(|(_, invite)| invite.user_id == user.id)
+                .map(|(code, invite)| (code.clone(), invite.clone()))
```
taggrx: Is there a good reason for cloning instead of returning references? If you struggle with life-times, let me know, I can rewrite this part.

src/backend/env/invite.rs
```diff
@@ -0,0 +1,138 @@
...
+                .collect::<Vec<_>>()
+        })
+        .unwrap_or_default()
+}
+
+pub fn validate_user_invites_credits(
```
taggrx, line 100: I (and I hope many) really would appreciate having explanatory docs on what this is doing, before digging into code.

src/backend/env/invite.rs
```diff
@@ -0,0 +1,138 @@
...
+            if total_with_diff > user.credits() {
+                return Err(format!(
+                    "You don't have enough credits to support invites {} , {}",
+                    user.credits(),
+                    total_with_diff
+                ));
+            }
+        }
+        None => {
+            return Err("Invite credits overflow".into());
```
taggrx: Can't we just handle it with `ok_or` in line 110?

src/backend/env/mod.rs
```diff
@@ -989,26 +1004,79 @@ impl State {
...
+    pub fn create_invite(
+        &mut self,
+        principal: Principal,
+        credits: Credits,
+        credits_per_user_opt: Option<Credits>,
+        realm_id: Option<RealmId>,
+    ) -> Result<(), String> {
+        let credits_per_user = credits_per_user_opt.unwrap_or(credits);
+        if credits % credits_per_user != 0 {
+            return Err("Credits per user are not a multiplier of credits".into());
```
taggrx:
```suggestion
            return Err("Credits per user are not a multiple of credits".into());
```

src/backend/env/mod.rs
```diff
@@ -989,26 +1004,79 @@ impl State {
...
             return Err(format!(
                 "smallest invite must contain {} credits",
                 min_credits
             ));
         }
-        if user.credits() < credits {
-            return Err("not enough credits".into());
-        }
+
+        invite::validate_realm_id(self, realm_id.as_ref())?;
```
taggrx: I think realm validation is the concern of the state module, so I think `validate_realm_id` could be just an private function on `state` itself.

src/backend/env/mod.rs
```diff
@@ -989,26 +1004,79 @@ impl State {
...
+            .ok_or(format!("Invite '{}' not found", invite_code))?;
+        if let Some(credits) = credits {
+            invite::validate_user_invites_credits(self, user, credits, Some(*invite_credits))?;
+
+            if credits % credits_per_user != 0 {
+                return Err(format!(
+                    "Credits per user are not a multiplier of credits {} {}",
+                    credits, credits_per_user
+                ));
+            }
```
taggrx: Shouldn't this be part of `validate_user_invites_credits`?

src/backend/env/mod.rs
```diff
@@ -989,26 +1004,79 @@ impl State {
...
+        }
+
+        self.invite_codes
+            .get_mut(&invite_code)
+            .ok_or(format!("Invite '{}' not found", invite_code))?
+            .update(credits, realm_id, user_id)?;
```
taggrx, line 1083: Would it make sense for `update` to run `validate_user_invites_credits` first? So that the caller doesn't need to think about validation? (this sounds more robust to me)

src/backend/env/user.rs
```diff
@@ -611,6 +611,21 @@ impl User {
...
         })
     }
+
+    pub fn validate_send_credits(&self, state: &State) -> Result<(), String> {
+        if let Some(invited_by) = self.invited_by {
+            let invite = state.invite_codes.values().find(|invite| {
+                invited_by == invite.user_id && invite.joined_user_ids.contains(&self.id)
+            });
+            if let Some(invite) = invite {
+                if self.credits_burned() < invite.credits_per_user && self.credits() < 1000 {
```
taggrx: The first condition is an excellent idea!!!

But why do we need the check `< 1000`?

Also, what's the life cycle of an invite? I think the old ones get deleted once used.

src/backend/env/invite.rs
aligatorr89: Yes, I like foreign key to be the same. Since invite has 2 relations to user, I changed it to `inviter_user_id`

src/backend/env/invite.rs
aligatorr89: Moved to the top. It might be better to panic which reverts state changes under mutate.

src/backend/env/user.rs
aligatorr89: Removed it. It was supposed to allow users that purchased credits, but attacker can simply send 1000 to all accounts and send 1050 back. 

Invite doesn't get removed anymore, it can be topped up when it reaches 0 credits. Use case will be:

- print qr code and paste on university/town-hall wall
- top it up based on how promotion is going

src/backend/env/invite.rs
aligatorr89: Yes, I've simplified it .

src/backend/env/mod.rs
aligatorr89: Do you think any function that uses state directly should be in mod.rs ?

src/backend/env/mod.rs
aligatorr89: validate_user_invites_credits uses state, while invite update only uses self (invite instance)

What should be done here in general? I've seen different approaches in codebase.

src/backend/env/invite.rs
aligatorr89: idea is that "" is allowed as a way to unset to None

src/backend/env/invite.rs
aligatorr89: Tbh, I was under the impression that mutate will auto-revert changes if result is err too. Thanks for spotting this!

src/backend/env/invite.rs
aligatorr89: It's done in create_invite in mod.rs . The reason is it uses global state.

src/backend/env/invite.rs
aligatorr89: I think it's nice to have. QR codes will be in phisical locations, so it's nice to have ability to change realm on them.

src/backend/env/mod.rs
aligatorr89: validate_user_invites_credits is based on all user invites in state. I moved it to invite::update

src/backend/env/invite.rs
taggrx: No, this in impossible to do in a general case, without cloning the whole state right?

src/backend/env/mod.rs
taggrx: Definitely not any, but realm validation sound like a useful function and can be probably used somewhere else too?

src/backend/env/invite.rs
taggrx: Hm, sorry I still don't understand, please explain with an example?

src/backend/env/invite.rs
taggrx: I think I do not understand the `Some("")` case at all yes, but I asked a related question on other thread.

src/backend/env/invite.rs
taggrx: When is such an invite expected to be clean up?

src/backend/env/mod.rs
taggrx: Oh I see the problem. Yeah, this is an ugly artifact of Rust... No, then let it as is.

src/backend/env/mod.rs
```diff
@@ -4843,6 +4912,46 @@ pub(crate) mod tests {
...
+            code
+        });
+
+        // New user should be joined to realm
+        let new_principal = pr(5);
+        assert_eq!(
+            State::create_user(new_principal, "name".to_string(), Some(invite_code)).await,
+            Ok(Some(realm_id.clone()))
+        );
+        read(|state| {
```
taggrx, line 4945: Should we also assert that user has a credit balance we expect and the invite's balance decreased as expected?

src/backend/env/invite.rs
aligatorr89: It will never clean up. We have validation against creating invites over balance, so it won't be infinite.

 We could cleanup invites that run out of credits and were not used in long time?

src/backend/env/invite.rs
aligatorr89: when you call update, you can unset it by "". Because of Option<> input, None means to ignore, while "" means to unset.

src/backend/env/mod.rs
aligatorr89: ofc, added

src/backend/env/mod.rs
aligatorr89: Yes, i moved it to mod.

src/backend/env/invite.rs
aligatorr89: Will refactor to Box, which is fixed size and more performant than Vec

src/backend/env/invite.rs
taggrx: Ok, maybe it's not urgent right now.

src/backend/env/invite.rs
taggrx: Correct, but we will never have `state.realms.contains_key(id)` being true in this case, right? So why the additional check for emptyness?

aligatorr89: Yes, but we want "" to skip check as it will be converted to None

aligatorr89: ah, I see, sec... :D

aligatorr89: Basically, we want "" to pass, cause it will be converted to None

src/backend/env/invite.rs
```diff
@@ -0,0 +1,143 @@
...
+
+impl Invite {
+    pub fn new(
+        credits: Credits,
+        credits_per_user: Credits,
+        realm_id: Option<RealmId>,
+        inviter_user_id: UserId,
+    ) -> Self {
+        // Convert empty realm_id to None
+        let converted_relm_id: Option<RealmId> = realm_id.filter(|id| !id.is_empty());
```
taggrx:
```suggestion
        let converted_realm_id: Option<RealmId> = realm_id.filter(|id| !id.is_empty());
```

src/backend/env/invite.rs
```diff
@@ -0,0 +1,143 @@
...
+
+        Self {
+            credits,
+            credits_per_user,
+            joined_user_ids: BTreeSet::new(),
+            realm_id: converted_relm_id,
```
taggrx:
```suggestion
            realm_id: converted_realm_id,
```

src/backend/env/invite.rs
```diff
@@ -0,0 +1,143 @@
...
+    ) -> Result<(), String> {
+        if self.inviter_user_id != user_id {
+            return Err("Owner does not match".into());
+        }
+
+        if let Some(new_credits) = credits {
+            if new_credits % self.credits_per_user != 0 {
+                return Err(format!(
+                    "Credits per user are not a multiple of credits {} {}",
+                    new_credits, self.credits_per_user
```
taggrx:
```suggestion
                    "Credits per user {} are not a multiple of new credits {}",
                    self.credits_per_user, new_credits,
```

src/backend/env/invite.rs
```diff
@@ -0,0 +1,143 @@
...
+        .map(|user_id| {
+            state
+                .invite_codes
+                .iter()
+                .filter(move |(_, invite)| invite.inviter_user_id == user_id)
+        })
+        .expect("Failed to filter invites");
+
+    Box::new(invites)
+}
```
taggrx: How about avoiding panicking?

```suggestion
pub fn invites_by_principal(
    state: &State,
    principal: Principal,
) -> Box<dyn Iterator<Item = (&'_ String, &'_ Invite)> + '_> {
    match state.principal_to_user(principal).map(|user| {
        state
            .invite_codes
            .iter()
            .filter(move |(_, invite)| invite.inviter_user_id == user.id)
    }) {
        Some(iter) => Box::new(iter),
        _ => Box::new(std::iter::empty()),
    }
}
```

src/backend/env/invite.rs
```diff
@@ -0,0 +1,143 @@
...
+
+/**
+ * Check allocated credits in invites do not exceed user's credits balance
+ *
+ * Protects against creating infinite number of invites
+ */
```
taggrx: Let's stick with the [RustDoc](https://doc.rust-lang.org/rustdoc/what-is-rustdoc.html) style?

```suggestion
/// Check allocated credits in invites do not exceed user's credits balance.
/// Protects against creating infinite number of invites.
```

src/backend/env/mod.rs
```diff
@@ -996,26 +1011,71 @@ impl State {
...
+        invite_code: String,
+        credits: Option<Credits>,
+        realm_id: Option<RealmId>,
+    ) -> Result<(), String> {
+        if credits.is_none() && realm_id.is_none() {
+            return Err("Update is empty".into());
+        }
+        let Some(user) = self.users.get(&user_id) else {
+            return Err("User not found".into());
+        };
```
taggrx:
```suggestion
        let user = self.users.get(&user_id).ok_or("user not found")?;
```

src/backend/env/invite.rs
```diff
@@ -0,0 +1,143 @@
...
+        .checked_add(new_credits)
+        .ok_or("Invite credits overflow")?
+        .checked_sub(old_credits.unwrap_or_default())
+        .ok_or("Invite credits overflow")?;
+
+    if total_with_diff > user.credits() {
+        return Err(format!(
+            "You don't have enough credits to support invites {} , {}",
+            user.credits(),
+            total_with_diff
```
taggrx:
```suggestion
            "not enough credits available: {} (needed for all open invites: {})",
            user.credits(),
            total_with_diff
```

src/backend/env/mod.rs
```diff
@@ -910,29 +916,50 @@ impl State {
...
+            if let Some((credits_per_user, inviter_id, code, realm_id)) =
+                invite_code.and_then(|code| {
+                    let invite = state.invite_codes.get(&code)?;
+                    Some((
+                        invite.credits_per_user,
+                        invite.inviter_user_id,
+                        code,
+                        invite.realm_id.clone(),
+                    ))
+                })
```
taggrx:
```suggestion
            if let Some((credits_per_user, inviter_id, code, realm_id)) =
                invite_code.and_then(|code| {
                    state.invite_codes.get(&code).map(|invite| {
                        (
                            invite.credits_per_user,
                            invite.inviter_user_id,
                            code,
                            invite.realm_id.clone(),
                        )
                    })
                })
```

src/backend/updates.rs
```diff
@@ -332,8 +305,18 @@ fn mint_credits() {
...
+    let (credits, credits_per_user, realm_id): (Credits, Option<Credits>, Option<RealmId>) =
+        parse(&arg_data_raw());
+    mutate(|state| reply(state.create_invite(caller(), credits, credits_per_user, realm_id)));
+}
+
+#[export_name = "canister_update update_invite"]
+fn update_invite() {
+    let (invite_code, credits, realm_id): (String, Option<Credits>, Option<RealmId>) =
+        parse(&arg_data_raw());
+    let user_id = read(|state| state.principal_to_user(caller()).expect("no user found").id);
```
taggrx: If I remember correctly this kind of app data is always extracted inside modules, not in `updates.rs` which mostly contains "plumbing" code.

Is there any reason to not simply pass `caller()` as a first argument and then find user and do all error handling inside `update_invite`?

src/frontend/src/invites.tsx
```diff
@@ -17,6 +35,56 @@ export const Invites = () => {
...
+            return;
+        }
+        return window.api
+            .call<any>(
+                "update_invite",
+                code,
+                updatedInvite.credits !== undefined &&
+                    updatedInvite.credits >= 0
+                    ? updatedInvite.credits
+                    : null,
```
taggrx: Wouldn't this work too here?
```suggestion
                updatedInvite.credits || null
```

src/frontend/src/invites.tsx
```diff
@@ -17,6 +35,56 @@ export const Invites = () => {
...
+                updatedInvite.realm_id !== undefined
+                    ? updatedInvite.realm_id
+                    : null,
```
taggrx: Same as above?

src/frontend/src/invites.tsx
```diff
@@ -17,6 +35,56 @@ export const Invites = () => {
...
+            )
+            .then(async (response) => {
+                if ("Err" in (response || {})) {
+                    alert(`Error: ${response.Err}`);
+                    setBusy(true);
+                    await loadInvites(); // Set back to prior state
+                    setBusy(false);
```
taggrx: Why do we need this? The state in UI should not have changed during the call, right?

src/frontend/src/invites.tsx
```diff
@@ -53,6 +124,22 @@ export const Invites = () => {
...
+                        }
+                    />
+                    Realm:
+                    <input
+                        type="text"
+                        value={inviteRealm}
+                        className="max_width_col"
+                        onChange={(event) => setInviteRealm(event.target.value)}
+                    />
                     {!busy && (
```
taggrx: I don't think we need this. Check out `ButtonWithLoading` instead. It simply takes an async function and displays loading while the function is called.

xqxpx (commented):
Please don't forget to lowercase all error messages 🙏

src/frontend/src/invites.tsx
aligatorr89: This is to reset to previous state if update fails.

src/frontend/src/invites.tsx
aligatorr89: 0 is valid number while null means empty.

src/backend/updates.rs
aligatorr89: Fully agreed, I changed it.

src/backend/env/invite.rs
aligatorr89: Much better, thanks!

aligatorr89:
> Please don't forget to lowercase all error messages 🙏

Done

xqxpx:
@aligatorr89 I added in 1332eb803e3d2b4b0897672bbf385404068e021e a quick test for the invite validation and instantly noticed a few things that could be improved.

Could you please also add a few tests for `consume` and `update` checking the edge cases?

src/backend/env/user.rs
```diff
@@ -611,6 +611,22 @@ impl User {
             Ok(())
         })
     }
+
+    /** Protect against invite phishing */
```
taggrx: Rustdoc please 😅

src/backend/env/user.rs
```diff
@@ -611,6 +611,22 @@ impl User {
...
+    pub fn validate_send_credits(&self, state: &State) -> Result<(), String> {
```
taggrx, line 618: We should add a todo here: since credits burned get reset every week, some users will never be able to transfer credits until the invite is alive. But we can handle this at some point later.

xqxpx (approved):

src/frontend/src/invites.tsx
```diff
@@ -17,95 +37,229 @@ export const Invites = () => {
...
+                alert(`Error: ${response.Err}`);
+                setBusy(true);
+                await loadInvites(); // Set back to prior state
+                setBusy(false);
+                return;
+            }
+        }
+    };
+
+    const updateInvite = (
```
taggrx, line 76: @aligatorr89 I changed the way invites are updated (modify in-place and set the `dirty` flag).

This is because I uncovered a bug: whenever I would update credits, the realm id would be deleted because the previous code would not send the realm id as updated value.

aligatorr89: Ok, I noticed, and added to tests

src/backend/env/user.rs
aligatorr89: Quite a surprise we don't keep the number. We could add a metric `credits_burned_total` which wouldn't reset?

PR #68, opened 2024-10-13
Halting

xqxpx:


aligatorr89 (approved):

ulan (commented):
Hey @taggrx. Shouldn't we update `State::init` to not overwrite balances from the transaction history? Otherwise, we would lose the balances?

ulan:
Maybe we should also clear `free` segments in the allocator such that the new data goes to a fresh memory? I am not sure.

ulan:
Oh, the balances field in `State` is marked as serde skip :(
So we are going to lose balances after upgrade.

ulan:
`User::balance` will be serialized though.

ulan:
Looks like we can restore balances of user accounts, but for principal that are not associated with users we need to restore manually from a snapshot that can be done now with
```
dfx canister call 6qfxa-ryaaa-aaaai-qbhsq-cai --ic balances --output=raw | xxd -r -p  > balances.json
```

ulan:
I have an idea that might work if we are lucky:

1. Get old transactions from the memory
2. Try to get recent transactions from the stable memory with a check that the transaction doesn't fall into the corrupted area. I.e. only load and deserialize the transaction if it is safe.
3. If there are no failed transactions, then it is safe to initialize balances from them.

xqxpx:
> Maybe we should also clear `free` segments in the allocator such that the new data goes to a fresh memory? I am not sure.

Now it would go into the first 56 MB of the stable memory which is the corrupted segment, so does it really matter?

xqxpx:
> Looks like we can restore balances of user accounts, but for principal that are not associated with users we need to restore manually from a snapshot that can be done now with
> 
> ```
> dfx canister call 6qfxa-ryaaa-aaaai-qbhsq-cai --ic balances --output=raw | xxd -r -p  > balances.json
> ```

Please see my OC message: we can recover all txs up until the migration to the stable memory: they are all in the old data structure, they were not destroyed (this was the next step).

And yes fetching the balances to reconcile later is a great idea!

xqxpx:
Also the e2e tests will not pass since the transfers are disabled, but I think this is fine.

ulan:
> Please see my OC message: we can recover all txs up until the migration to the stable memory: they are all in the old data structure, they were not destroyed (this was the next step).

Yes, I understand that. Transactions since Sep 27 will be lost, right? Balances will be also reverted to Sep 27 after upgrade.

What is the exact plan to restore the correct balances? Let's think this through before doing the upgrade that destroys the balances.

Here is my thinking:
1. We ask exchanges to halt trading of TAGGR.
2. We ask everyone who is interested to download a snapshot of balances.
3. There is consensus on that snapshot and we are going to restore balances to that snapshot.
4. We hardcode the snapshot of balances into post-upgrade such that after the upgrade we restore balances to the snapshot that everyone agreed on.
5. We do the upgrade.
6. Now balances are correct, but transactions since Sep 27 are missing.
7. We download the stable memory and try to recover the missing transactions from there?
8. If we fail to recover the missing transaction, we reconcile by adding minting/burning transactions such that the transaction history agrees with the balances?

Do you have something else in mind?

ulan:
> I have an idea that might work if we are lucky:
> 
> 1. Get old transactions from the memory
> 2. Try to get recent transactions from the stable memory with a check that the transaction doesn't fall into the corrupted area. I.e. only load and deserialize the transaction if it is safe.
> 3. If there are no failed transactions, then it is safe to initialize balances from them.

My idea doesn't work because the Api state about locations of transactions in stable memory was reset in heap. However, @taggrx you should have that data in the very first backup that you did. If you still have that backup, please keep it, it might be useful for recovering the missing transactions.

xqxpx:
> Yes, I understand that. Transactions since Sep 27 will be lost, right? 

Yes, this is why we needed the balances.

> 1. We ask exchanges to halt trading of TAGGR.
> 2. We ask everyone who is interested to download a snapshot of balances.
> 3. There is consensus on that snapshot and we are going to restore balances to that snapshot.
> 4. We hardcode the snapshot of balances into post-upgrade such that after the upgrade we restore balances to the snapshot that everyone agreed on.
> 5. We do the upgrade.
> 6. Now balances are correct, but transactions since Sep 27 are missing.
> 7. We download the stable memory and try to recover the missing transactions from there?
> 8. If we fail to recover the missing transaction, we reconcile by adding minting/burning transactions such that the transaction history agrees with the balances?
> 
> Do you have something else in mind?

This plan sounds great event though I am not sure if we need to recover the missing TXs. A reconciling with balances should be enough. The missing txs were not fetched by any 3rd party, so why do we want to care about them?


And btw one missing piece: I think this PR needs another code: recompute balances from the in-memory ledger, so that we can keep upgrading later! Right?

I would also maybe add a couple of error handling so that the interface can be loaded and we can upgrade via proper interface.

xqxpx:
Pushed new commits: 

- let the UI load (then people also have access to their ICP wallets and upgrade proposals).
- disable any modifications of the stable memory
- disable transfers
- disable panicking on stable memory reading (this will return default nonce data)

src/backend/env/memory.rs
```diff
@@ -362,18 +362,13 @@ impl<K: Ord + Eq, T: Serialize + DeserializeOwned> Default for ObjectManager<K,
...
     }
 
     pub fn insert(&mut self, id: K, value: T) -> Result<(), String> {
-        assert!(self.initialized, "allocator uninitialized");
-        if self.index.contains_key(&id) {
-            self.remove(&id)?;
-        }
-        self.index.insert(id, self.api.borrow_mut().write(&value)?);
-        Ok(())
+        unimplemented!()
```
ulan: Are you sure this is not going to trigger a panic in the pre-upgrade? Or prevent emergency upgrades somehow? I cannot foresee the consequences and side-effects of this change, so it seems dangerous to me.

src/backend/env/memory.rs
```diff
@@ -393,11 +388,7 @@ impl<K: Eq + Ord + Clone + Display, T: Serialize + DeserializeOwned> ObjectManag
     }
 
     pub fn remove(&mut self, id: &K) -> Result<T, String> {
-        assert!(self.initialized, "allocator uninitialized");
-        let (offset, len) = self.index.remove(id).ok_or("not found")?;
-        let value = self.api.borrow().read(offset, len);
-        self.api.borrow_mut().remove(offset, len)?;
-        Ok(value)
+        unimplemented!()
```
ulan: ditto

src/backend/env/mod.rs
```diff
@@ -579,7 +579,9 @@ impl State {
 
     pub fn init(&mut self) {
         assets::load();
-        match token::balances_from_ledger(&mut self.memory.ledger.iter().map(|(_, tx)| tx)) {
+        match token::balances_from_ledger(
```
ulan: If we go with the plan I proposed, we would need to embed the JSON file with balances and check that its hash matches the "decentralized consensus" value. Then load the balances from that JSON.

ulan:
To me it seems safer to keep the UI non-functional until we fully restore as much as we can. Having a semi-working UI increases chances of unexpected operations done by users.

ulan:
> The missing txs were not fetched by any 3rd party, so why do we want to care about them?

I think it is important to retain the correct history. At least I can try and if I fail, then we could reconcile with minting/burning.

ulan:
> And btw one missing piece: I think this PR needs another code: recompute balances from the in-memory ledger, so that we can keep upgrading later! Right?

If we go with the decentralized snapshot of balances, then we should restore directly from it.

ulan:
@taggrx do you have a backup on Sep 27? Also do you have the first backup from yesterday? Could you share them somehow?

ulan:
How about changing the allocator such that it initializes its boundary to the end of stable memory? So that all new allocation happen in the newly allocator memory and don't overwrite existing data?

ulan:
Then we don't have to disallow writing into the stable memory.

ulan:
I am going afk for 6-7 hours. I'll check the messages in the evening. Let's not rush this and think through as much as we can.

src/backend/env/memory.rs
taggrx: It should not, but at the same time we can simply do nothing, not panic. Will change!

src/backend/env/mod.rs
taggrx: Ah so you want to do it all in the same release? But how can we prevent balance changes between the upgrade and proposal with a json file?

xqxpx:
I only have a backup from 27th. Yesterday, when I ran the backup it didn't work anymore, the heap was already at the start.

I'll try to find a way to share it.

xqxpx:
> Then we don't have to disallow writing into the stable memory.

I'm not sure what's riskier. If UI is disabled and the cold archiving is disabled, there will be no write operations to the stable memory anyway.

src/backend/updates.rs
```diff
@@ -83,30 +87,22 @@ fn post_upgrade() {
...
+        assert_eq!(
+            format!("{:x}", result),
+            "7e3752fa3ca60e2af537e5bb81179128c9ef335270bd84fa13d3760f9a619c25"
+        );
+
+        let records: Vec<(Account, u64, Option<UserId>)> =
+            serde_json::from_slice(&balances).unwrap();
+
+        for r in records {
+            state.balances.insert(r.0, r.1);
```
ulan, line 104: Shouldn't we also update `User` balances here?

src/backend/updates.rs
```diff
@@ -83,30 +87,22 @@ fn post_upgrade() {
...
-
-            // Move all txs to stable memory
-            for tx in state.ledger.iter() {
-                state
-                    .memory
-                    .ledger
-                    .insert(tx.id as u32, tx.clone())
-                    .unwrap();
-            }
-            state.init();
```
ulan, line 109: Should the code in `State::init` that updates balances from the transaction history be also disabled?

ulan: Nevermind, I missed the change.

src/backend/updates.rs
taggrx: Oh yes thanks!

ulan (approved):
LGTM

chmllr:
Changes lgtm 👍

PR #81, opened 2025-01-09
Feature request: Add ability to manage Icrc1 tokens

aligatorr89:
Ability to manage tokens in frontend. Tokens are persisted for great UX.

image

Mobile:

image


src/frontend/src/api.ts
```diff
@@ -118,6 +122,18 @@ export const ApiGenerator = (
             console.error(err);
         });
 
+    const icrc_canisters: Map<string, IcrcLedgerCanister> = new Map();
+    const getIcrcCanister = (canisterId: string) => {
+        const canisterAgent = icrc_canisters.get(canisterId);
```
taggrx: We never add anything to `icrc_canisters`, do we?

src/frontend/src/api.ts
```diff
@@ -483,16 +499,61 @@ export const ApiGenerator = (
...
             return await canister.balance({
                 certified: false,
                 owner: account.owner,
                 subaccount: account.subaccount,
             });
         },
+        icrc_metadata: async (canisterId: string) => {
+            try {
+                const canister = getIcrcCanister(canisterId);
+                // console.log(canister.metadata)
```
taggrx: Leftover

src/frontend/src/util.ts
```diff
@@ -0,0 +1,20 @@
+import { MAINNET_MODE } from "./env";
+
+export function bucket_image_url(
```
taggrx: We have already `common.ts`, no need for a new file.

src/frontend/src/icrc1-tokens-wallet.tsx
```diff
@@ -0,0 +1,226 @@
...
+                                            )
+                                        }
+                                        label={"Send"}
+                                    ></ButtonWithLoading>
+                                </td>
+                                <td>
+                                    <span className="monospace">{`${(+canisterBalances[canisterId] / Math.pow(10, info.decimals))?.toFixed(info.decimals)}`}</span>
+                                </td>
+                                <td>
+                                    <CopyToClipboard
```
taggrx: Why do you think we need a Receive button copying exactly the same value, if we have it once already both for TAGGR and ICP at the top of the bottom?

xqxpx:
Do you mind if I work on top of this PR a bit to consolidate the TAGGR tokens UI section with the new wallet?

aligatorr89:
> Do you mind if I work on top of this PR a bit to consolidate the TAGGR tokens UI section with the new wallet?

not at all :) . Let me know if you need any help or want to split something.

src/frontend/src/icrc1-tokens-wallet.tsx
aligatorr89: I saw it in other wallets, and most people don't understand it's same value :D

src/frontend/src/api.ts
aligatorr89: To avoid instantiating it on every call.

src/frontend/src/api.ts
aligatorr89: cheers, I guess you will fix with your changes ?

src/frontend/src/icrc1-tokens-wallet.tsx
taggrx: I'd be reactive here, not pro-active. 

E.g. in Plug the account / principal id are displayed once as well, I think it is a common understanding.

src/frontend/src/api.ts
taggrx: I understand, but then we should actually add something after the very first instantion?

xqxpx:
@aligatorr89 pushed changes, however the persistence of the token lists is in browser only... this is a recipe for endless enquiries from users :) I think it should be simply stored on the user object. Can you please add this?

aligatorr89:
> @aligatorr89 pushed changes, however the persistence of the token lists is in browser only... this is a recipe for endless enquiries from users :) I think it should be simply stored on the user object. Can you please add this?

I will add canister persistence. Thanks for your invaluable fixes!

aligatorr89:
> Do you mind if I work on top of this PR a bit to consolidate the TAGGR tokens UI section with the new wallet?

aligatorr89:
> @aligatorr89 pushed changes, however the persistence of the token lists is in browser only... this is a recipe for endless enquiries from users :) I think it should be simply stored on the user object. Can you please add this?

~I added to `update_user` mutation, which is big and can cause issues longterm, because parameters are overwritten. Let me know if it's ok.~

added `update_user_icrc_canisters` for simplicity

xqxpx:
@aligatorr89 do you need help with the e2e test?

aligatorr89:
> @aligatorr89 do you need help with the e2e test?

hey @taggrx yes, I don't know where it broke. Maybe actions evnironment have diverged. Additionally we are facing build issue.

PR #83, opened 2025-02-15
fix: backup and restore tool

6qfxa:
The existing backup tool returns errors:
```
Fetching page 0...
Status endpoint returned an invalid status.
Invalid JSON content
```

It could be caused by the old version of ic-agent. Updating cargo dependencies is not easy due to breaking changes.

It is easier to rewrite the backup tool to directly call the canister backup/restore methods with the new version ic-agent.

The rewrite removes 650 lines of code, adds 170 lines of code (not counting Cargo.lock changes)

The existing LICENSE file is kept intact, but could be removed in favor of the top-level LICENSE
because the new code does not use code from `qu`.


6qfxa:
@taggrx: if backup fails for you too, then this is a fix ^

xqxpx (approved):
Awesome patch! Thanks, @6qfxa and welcome to anonymity!

6qfxa:
@taggrx: thanks, I assume you will merge it? The test failure appears unrelated to it.

PR #86, opened 2025-03-08
Fix failing E2E test: test1.spec.ts/Wallet

6qfxa:
Setup.ts of E2E test calls `dfx canister call taggr reset` that fetches an external xdr rate.

This patch initializes the xdr rate to a fixed value to make tests less fragile.

### DIFF
--- a/src/backend/env/mod.rs
+++ b/src/backend/env/mod.rs
@@ -1728,6 +1728,17 @@ impl State {
         }
     }
 
+    pub fn get_xdr_rate() -> u64 {
+        read(|state| state.e8s_for_one_xdr)
+    }
+
+    pub fn reset_xdr_rate_for_testing() {
+        mutate(|state| {
+            // If this is set to 1_000_000 and above, then E2E tests fail.
+            state.e8s_for_one_xdr = 900_000;
+        });
+    }
+
     pub async fn hourly_chores(now: u64) {
         mutate(|state| {
             state.backup_exists = false;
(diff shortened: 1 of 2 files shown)


6qfxa:
@aligatorr89, @taggrx: please take a look.

src/backend/env/mod.rs
```diff
@@ -1728,6 +1728,17 @@ impl State {
         }
     }
 
+    pub fn get_xdr_rate() -> u64 {
+        read(|state| state.e8s_for_one_xdr)
+    }
+
+    pub fn reset_xdr_rate_for_testing() {
```
aligatorr89, line 1736: Add protection for compiling only in test or dev

```suggestion
    #[cfg(any(test, feature = "dev"))]
    pub fn reset_xdr_rate_for_testing() {
```

aligatorr89 (changes requested):

aligatorr89 (approved):

aligatorr89:
Great find @6qfxa 🎉

xqxpx:
Thanks so much! But do we have an explanation why the test broke? The rate was always fetched from the canisters deployed during the test. Maybe the XDR rate canister started serving a different rate?

aligatorr89:
@6qfxa what is your Taggr ticker ? Would you be interested in reviewing external tipping ? https://github.com/TaggrNetwork/Taggr/pull/87

PR #87, opened 2025-03-21
Realm native token and tipping

aligatorr89:
Realm native token and tipping

- add native token and more tokens to realm
- tip with realm token sand show it on Taggr posts
- e2e tests, rust unit tests
- integrate IcExplorer for auto discovery of tokens in wallet - UX improvement

Workflow for tips:
1. tip client -> external ledger
2. add to the view optimistically (good UX and quick)
3. verify tip with `icrc3_get_blocks` and add it to post, cost is 5 credits to cover for canister->canister call

For next time - min_native_token_balance:
- provide a funding model for DAO - canister -> canister calls burn cycles
- we can increase post costs to fund it
- if below min, we remove post -> more credits burned to disinsentivize

GEMINI review:
[taggr-review.md](https://github.com/user-attachments/files/22083666/taggr-review.md)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,9 +2,9 @@ start:
 	ulimit -n 65000 && dfx start --background -qqqq 2>&1 | grep -v sgymv &
 
 staging_deploy:
-	NODE_ENV=production DFX_NETWORK=staging make fe
+	NODE_ENV=production DFX_NETWORK=$(if $(CANISTER),$(CANISTER),staging) make fe
 	FEATURES=staging dfx build
-	FEATURES=staging dfx --identity prod deploy --network staging taggr
+	FEATURES=staging dfx --identity prod deploy --network $(if $(CANISTER),$(CANISTER),staging) taggr
 
 local_deploy:
 	FEATURES=dev dfx deploy
--- a/e2e/test1.spec.ts
+++ b/e2e/test1.spec.ts
@@ -286,8 +286,11 @@ test.describe("Upgrades & token transfer flow", () => {
         test("Add - input", async () => {
             // Enable in settings
             await page.goto("/#/settings");
-            await expect(page.getByTestId("ic-wallet-select")).toBeVisible();
-            await page.getByTestId("ic-wallet-select").selectOption("YES");
+            const icrcWalletEnableSelect = page.getByTestId("ic-wallet-select");
+            await expect(icrcWalletEnableSelect).toBeVisible();
+            await icrcWalletEnableSelect.selectOption("YES");
+            await icrcWalletEnableSelect.selectOption("YES"); // Select twice due to a bug in UI
+
             await page.getByRole("button", { name: "SAVE" }).click();
             await page.waitForTimeout(1000);
             await page.reload();
--- a/e2e/test1.spec.ts
+++ b/e2e/test1.spec.ts
@@ -1,10 +1,44 @@
-import { test, expect, Page } from "@playwright/test";
+import { test, expect, Page, Locator } from "@playwright/test";
 import { resolve } from "node:path";
 import { exec, mkPwd } from "./command";
 import { createHash } from "node:crypto";
 import { readFile } from "node:fs/promises";
 
 test.describe.configure({ mode: "serial" });
+
+const executeTransfer = async (page: Page, btn: Locator, amount = "5") => {
+    // Q1 - enter the principal receiver
+    await new Promise((resolve) => {
+        page.once("dialog", async (dialog) => {
+            if (dialog.message().includes("Enter the recipient principal")) {
+                await dialog.accept("6qfxa-ryaaa-aaaai-qbhsq-cai");
+                resolve(null);
+            }
+        });
+        // Click button after listener is setup
+        btn.click();
+    });
+
+    // Q2 - enter the amount
+    await new Promise((resolve) => {
+        page.once("dialog", async (dialog) => {
+            if (dialog.message().includes("Enter the amount")) {
+                await dialog.accept(amount);
+                resolve(null);
+            }
+        });
+    });
+
+    // Q3 - confirm receiver and amount
+    await new Promise((resolve) => {
+        page.once("dialog", async (dialog) => {
+            await dialog.accept();
+            await page.waitForLoadState("networkidle");
+            await page.waitForTimeout(1000);
+            resolve(null);
+        });
+    });
+};
 
 test.describe("Upgrades & token transfer flow", () => {
     let page: Page;
@@ -73,11 +107,7 @@ test.describe("Upgrades & token transfer flow", () => {
             .getByTestId("reaction-picker")
             .click();
         // React with a star
-        await page
-            .locator('button[title="Reward points: 10"]')
-            .first()
-            .click({ delay: 3000 });
-        await page.waitForTimeout(4500);
+        await page.locator('button[title="Reward points: 10"]').first().click();
     });
 
     test("Create an auction bid, trigger minting", async ({}) => {
@@ -101,28 +131,7 @@ test.describe("Upgrades & token transfer flow", () => {
 
         await expect(page.getByTestId("token-balance")).toHaveText("15");
 
-        const transferExecuted = new Promise((resolve, _reject) => {
-            page.on("dialog", async (dialog) => {
-                if (
-                    dialog.message().includes("Enter the recipient principal")
-                ) {
-                    await dialog.accept("6qfxa-ryaaa-aaaai-qbhsq-cai");
-                }
-                if (dialog.message().includes("Enter the amount")) {
-                    await dialog.accept("5");
-                }
-                if (dialog.message().includes("You are transferring")) {
-                    await dialog.accept();
-                    await page.waitForLoadState("networkidle");
-                    await page.waitForTimeout(3000);
-                    resolve(null);
-                }
-            });
-        });
-
-        await page.getByTestId("tokens-transfer-button").click();
-
-        await transferExecuted;
+        await executeTransfer(page, page.getByTestId("tokens-transfer-button"));
 
         await expect(page.getByTestId("token-balance")).toHaveText("9.9");
         await page.getByTestId("token-balance").click();
@@ -271,6 +280,117 @@ test.describe("Upgrades & token transfer flow", () => {
         expect(
             await page.locator("p", { hasText: /Upgrade succeeded/ }).count(),
         ).toEqual(2);
+    });
+
+    test.describe("IC TOKENS", () => {
+        test("Add - input", async () => {
+            // Enable in settings
+            await page.goto("/#/settings");
+            await expect(page.getByTestId("ic-wallet-select")).toBeVisible();
+            await page.getByTestId("ic-wallet-select").selectOption("YES");
+            await page.getByRole("button", { name: "SAVE" }).click();
+            await page.waitForTimeout(1000);
+            await page.reload();
+
+            // Test the wallet functionality
+            await page.goto("/");
+            await page.getByTestId("toggle-user-section").click();
+
+            await expect(page.getByTestId("token-balance")).toHaveText("9.9"); // Starting balance
+
+            const tokenListLocator = page.locator(
+                'div[data-testid="ic-tokens-div"]',
+            ); // IC tokens list locator
+            expect(tokenListLocator).toBeVisible();
+
+            const addTokenDialog = new Promise((resolve, _reject) => {
+                page.once("dialog", async (dialog) => {
+                    if (dialog.message().includes("ICRC canister id")) {
+                        await dialog.accept("bkyz2-fmaaa-aaaaa-qaaaq-cai"); // Add Taggr token
+                        resolve(null);
+                    }
+                });
+            });
+
+            const addTokenBtn = tokenListLocator.locator(
+                'button[title="Add token"]',
+            );
+            await expect(addTokenBtn).toBeEnabled();
+            await addTokenBtn.click();
+
+            // Add token and see it in the list
+            await addTokenDialog;
+
+            await expect(
+                page.getByTestId("bkyz2-fmaaa-aaaaa-qaaaq-cai-balance"),
+            ).toHaveText("9.90"); // Token added to the list
+        });
+
+        test("Send", async () => {
+            const tokenSendBtn = page.getByTestId(
+                "bkyz2-fmaaa-aaaaa-qaaaq-cai-send",
+            );
+
+            await expect(tokenSendBtn).toBeVisible();
+
+            await executeTransfer(page, tokenSendBtn); // Token send button
+
+            await expect(
+                page.getByTestId("bkyz2-fmaaa-aaaaa-qaaaq-cai-balance"),
+            ).toHaveText("4.80"); // Starting balance
+        });
+
+        test("Hide zeros", async () => {
+            const tokenHideZeros = page.getByTestId(
+                "canisters-hide-zero-balance",
+            );
+            const tokenSendBtn = page.getByTestId(
+                "bkyz2-fmaaa-aaaaa-qaaaq-cai-send",
+            );
+
+            await executeTransfer(page, tokenSendBtn, "4.70"); // Send all remaining balance
+            await expect(
+                page.getByTestId("bkyz2-fmaaa-aaaaa-qaaaq-cai-balance"),
+            ).toHaveText("0.00"); // 0 balance
+
+            // Hide zeros
+            await expect(tokenHideZeros).toBeVisible();
+            await tokenHideZeros.click();
+
+            // Token removed from the list
+            await expect(
+                page.getByTestId("bkyz2-fmaaa-aaaaa-qaaaq-cai-send"),
+            ).not.toBeVisible();
+
+            await tokenHideZeros.click(); // Add zeros back to the list
+            await expect(
+                page.getByTestId("bkyz2-fmaaa-aaaaa-qaaaq-cai-send"),
+            ).toBeVisible();
+        });
+
+        test("Remove", async () => {
+            const tokenRemoveBtn = page.getByTestId(
+                "bkyz2-fmaaa-aaaaa-qaaaq-cai-remove",
+            );
+
+            await expect(tokenRemoveBtn).toBeVisible();
+
+            const removeTokenDialog = new Promise((resolve, _reject) => {
+                page.once("dialog", async (dialog) => {
+                    if (dialog.message().includes("Remove TAGGR")) {
+                        await dialog.accept(); // Confirm remove token
+                        await page.waitForTimeout(1000);
+                        resolve(null);
+                    }
+                });
+            });
+
+            await tokenRemoveBtn.click(); // Remove token
+            await removeTokenDialog;
+            expect(
+                page.getByTestId("bkyz2-fmaaa-aaaaa-qaaaq-cai-remove"),
+            ).not.toBeVisible(); // Row removed
+        });
     });
 });
 
--- a/e2e/test3.spec.ts
+++ b/e2e/test3.spec.ts
@@ -176,9 +176,9 @@ test.describe("Regular users flow, part two", () => {
             await expect(postMenu).toBeVisible();
             await postMenu.locator(`button[title="Tip"]`).click();
             // Wait for custom popup and send 1 Taggr
-            const popup = page.getByTestId("custom-popup");
+            const popup = page.getByTestId("popup");
             await expect(popup).toBeVisible();
-            await expect(popup).toHaveText(/Tip @john with.*/);
+            await expect(popup).toHaveText(/Tip john with.*/);
             await popup.locator("input").fill("1"); // Send 1 Taggr to john
 
             popup.getByText("SEND").click();
@@ -216,7 +216,7 @@ test.describe("Regular users flow, part two", () => {
             await expect(postMenu).toBeVisible();
             await postMenu.locator(`button[title="Tip"]`).click();
             // Wait for custom popup and send 1 Taggr
-            const popup = page.getByTestId("custom-popup");
+            const popup = page.getByTestId("popup");
             await expect(popup).toBeVisible();
             await popup.locator("input").fill("1"); // Send 1 Taggr to john
             popup.getByText("SEND").click();
--- a/e2e/test3.spec.ts
+++ b/e2e/test3.spec.ts
@@ -105,4 +105,134 @@ test.describe("Regular users flow, part two", () => {
         });
         await expect(repostFeedItem.getByText(/Poll from John/)).toBeVisible();
     });
+
+    test.describe("Tips", () => {
+        test('Logout and login with "eye" user', async () => {
+            // Logout and register "eye" user
+            await page.getByTestId("toggle-user-section").click();
+            await expect(page.locator(`a[title="SIGN OUT"]`)).toBeVisible();
+            await page.locator(`a[title="SIGN OUT"]`).click();
+
+            await expect(page).toHaveTitle("TAGGR");
+
+            await page.getByRole("button", { name: "SIGN UP" }).click();
+            await page.getByRole("button", { name: "SEED PHRASE" }).click();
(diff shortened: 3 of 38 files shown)


Makefile
```diff
@@ -6,6 +6,11 @@ staging_deploy:
 	FEATURES=staging dfx build
 	FEATURES=staging dfx --identity prod deploy --network staging taggr
 
+staging_dev2_deploy:
```
taggrx: Instead of adding a new command, let's parameterize the existing `staging_deploy` like how I parameterized `make release`. E.g. I'd rather suggest to deploy to the selected staging by something like:

CANISTER=staging_dev2 make staging_deploy

taggrx: Another comment I have is what does `dev` in the name of the second staging environment imply? Staging is obviously dev, not prod. How about simply `staging2`?

e2e/test1.spec.ts
```diff
@@ -1,11 +1,45 @@
...
+const executeTransfer = async (page: Page, btn: Locator, amount = "5") => {
+    // Q1 - enter the principal receiver
+    await new Promise((resolve) => {
+        page.once("dialog", async (dialog) => {
+            if (dialog.message().includes("Enter the recipient principal")) {
+                await dialog.accept("6qfxa-ryaaa-aaaai-qbhsq-cai");
+                resolve(null);
+            }
+        });
+        // Click button after lisenter is setup
```
taggrx:
```suggestion
        // Click button after listener is setup
```

e2e/test1.spec.ts
```diff
@@ -265,6 +274,117 @@ test.describe("Upgrades & token transfer flow", () => {
             await page.locator("p", { hasText: /Upgrade succeeded/ }).count(),
         ).toEqual(2);
     });
+
+    test.describe("IC TOKENS", () => {
```
taggrx, line 285: Awesome! Thank you!

e2e/test3.spec.ts
```diff
@@ -104,4 +104,128 @@ test.describe("Regular users flow, part two", () => {
         });
         await expect(repostFeedItem.getByText(/Poll from John/)).toBeVisible();
     });
+
+    test.describe("Tips", () => {
+        test('Logout and login with "1" user', async () => {
```
taggrx: Why not using a random name? I had to read the code to understand what "1" user means...

src/backend/env/mod.rs
```diff
@@ -145,6 +147,11 @@ pub struct Realm {
     pub posts: Vec<PostId>,
     pub adult_content: bool,
     pub comments_filtering: bool,
+    /// Native token of realm - 1 use case is min balance to post
```
taggrx: I would describe use cases etc in the white paper.

```suggestion
    /// Native token of the realm
```

src/backend/env/mod.rs
```diff
@@ -778,6 +792,9 @@ impl State {
         if !logo.is_empty() {
             realm.logo = logo;
         }
+        if tokens.clone().map_or(false, |t| t.len() > 50) {
```
taggrx: We shouldn't clone during the validation, `.as_ref()` should be fine here.

src/backend/env/mod.rs
```diff
@@ -778,6 +792,9 @@ impl State {
...
+            return Err("tokens count below 50".into());
```
taggrx:
```suggestion
            return Err("tokens count above 50".into());
```

src/backend/env/post.rs
```diff
@@ -91,6 +91,9 @@ pub struct Post {
 
     #[serde(default)]
     pub encrypted: bool,
+
+    #[serde(default)]
+    pub has_external_tip: Option<bool>,
```
taggrx: Please add comments explaining what it is. For example, I didn't see the rest of the code, but seeing this with this name, I have no idea why this field is needed.

src/backend/env/tip.rs
```diff
@@ -0,0 +1,385 @@
...
+#[derive(Clone, Serialize, Deserialize)]
+pub struct Tip {
+    id: TipId,
+    post_id: PostId,
+
+    sender_id: UserId,
+
+    canister_id: Principal,
+    amount: u128,
+    /// Index from external transaction
```
taggrx, line 15: Why is it needed?

src/backend/env/tip.rs
```diff
@@ -0,0 +1,385 @@
...
+    state: &mut State,
+    post_id: PostId,
+    canister_id: Principal,
+    amount: u128,
+    memo: Option<Vec<u8>>,
+    to_principal: Principal,
+    from_principal: Principal,
+    index: u64,
+) -> Result<Tip, String> {
+    let post = Post::get(state, &post_id).ok_or("post not found")?;
```
taggrx, line 31: Nit: reading the post is the most expensive operation here, I'd do it after we know we have all user accounts, so move it down as much as possible

src/backend/env/tip.rs
```diff
@@ -0,0 +1,385 @@
...
+    } else {
+        state.post_tip_indexes.insert(post_id, vec![tip.id]);
+    }
+
+    state
+        .users
+        .get_mut(&receiver_id)
+        .expect("user not found")
+        .notify_about_post(
+            format!("@{} tipped you with for your post", sender_name),
```
taggrx: Something is missing in the message

src/backend/env/tip.rs
```diff
@@ -0,0 +1,385 @@
...
+            post_id,
+        );
+
+    Ok(tip)
+}
+
+fn convert_memo_to_u64(memo: Option<Vec<u8>>) -> Result<PostId, String> {
```
taggrx: Is this really a `PostId`?

src/backend/env/tip.rs
```diff
@@ -0,0 +1,385 @@
...
+    }
+
+    let mut num = 0u64;
+
+    // Combine bytes in little-endian order
+    for &byte in vec.iter() {
+        num = (num << 8) | byte as u64; // Shift left and add byte
+    }
+
+    Ok(num)
```
taggrx: I didn't test it, but this is more idiomatic, please TAL:
```suggestion
let mut padded = [0u8; 8];
let len = std::cmp::min(memo_value.len(), 8);
padded[8-len..].copy_from_slice(&memo_value[..len]);
Ok(u64::from_be_bytes(padded))
```

e2e/test1.spec.ts
aligatorr89: awesome!

src/backend/env/tip.rs
aligatorr89: - to avoid repeating same tip, only way to make sure it's not repeated
- generally it's good idea to keep externalId

src/backend/env/tip.rs
aligatorr89: u64 obviouselly :D

src/backend/env/tip.rs
aligatorr89: amazing, works the same. How did you spot this ?

src/backend/env/mod.rs
aligatorr89: removed native token due to reasons described

Makefile
aligatorr89: done, original staging is default

src/backend/env/tip.rs
aligatorr89: We don't know symbol and decimals in backend. User can still see it in notification by clicking 

image

src/backend/env/canisters.rs
```diff
@@ -251,6 +289,20 @@ pub async fn coins_for_one_xdr(coin: &str) -> Result<u64, String> {
         .map(|result| result.rate / 10)
 }
 
+pub async fn icrc_transactions(
+    canister_id: Principal,
+    args: GetTransactionsArgs,
+) -> Result<GetTransactionsResult, String> {
+    let (response,): (GetTransactionsResult,) =
+        call_canister(canister_id, "get_transactions", (args,))
```
taggrx: But that's not an ICRC standard.

src/backend/env/mod.rs
```diff
@@ -145,6 +147,8 @@ pub struct Realm {
     pub posts: Vec<PostId>,
     pub adult_content: bool,
     pub comments_filtering: bool,
+    /// Tokens allowed to appear in realm like tips
+    pub tokens: Option<BTreeSet<Principal>>,
```
taggrx: Did we say we do not white list tip tokens?

src/backend/env/mod.rs
```diff
@@ -262,6 +266,10 @@ pub struct State {
     pub weekly_chores_delay_votes: HashSet<UserId>,
 
     pub timers: Timers,
+
+    // Map of post id to sorted of tip ids and external index
+    #[serde(default)]
+    pub post_tip_indexes: BTreeMap<PostId, Vec<TipId>>,
```
taggrx: Why is this data structure needed? Why can't all tip data be on the post itself?

src/backend/updates.rs
```diff
@@ -624,6 +627,64 @@ fn cancel_bid() {
     spawn(async { reply(auction::cancel_bid(caller()).await) });
 }
 
+#[export_name = "canister_update add_external_icrc_transaction"]
+fn add_external_icrc_transaction() {
```
taggrx, line 706: I would move all actual logic to the tip.rs module, to keep the update handler  tidy.

src/backend/updates.rs
```diff
@@ -624,6 +627,64 @@ fn cancel_bid() {
...
+    let (canister_id_as_str, start_index, post_id): (String, u64, PostId) = parse(&arg_data_raw());
```
taggrx, line 707: I don't understand... Looks like I can call this function with any post_id and _any_ start_index I want (even if I didn't make it) and this will be just added as a tip to that post?!

aligatorr89: if you want to burn 5 credits for nothing yes :)

src/backend/env/mod.rs
aligatorr89: Solid design, 1-many table. For Taggr posts, it's wasteful to load when it's never shown unless clicked on post menu.

src/backend/env/mod.rs
aligatorr89: realm controllers want this, since it gives just few options they want

src/backend/env/canisters.rs
aligatorr89: It's Rosetta standard, all ledgers seem to have it. It holds only recent transactions by index, excatly what we need here. Thanks for spotting, renamed to `get_transactions

src/backend/updates.rs
aligatorr89: good idea

src/backend/env/mod.rs
taggrx: This is wrong: if it is attached to the post, you don't need to store the postID separately and en empty vector is a dangling pointer which is allocates 0 bytes! With the global table, you introduce a whole new structure in the persistent storage and store each tip separately, which is also conceptually weird to explain to others: why storing parts of the posts separately? And since the persistent storage always stores data in blocks, you almost always allocate a couple of bytes that you'll never use. This is what is really wasteful for no good reason, IMO.

src/backend/updates.rs
taggrx: Wait, but if we think it's ok to attach any untrusted transaction to the post, why do even have any logic in the backend (`tip.rs`)? We could simply make an ICRC transaction on the frontends and simply call this function with all the parameters? Why are we making then an expensive inter-canister call to fetch data that is not verified?

taggrx: Also, I am not sure in general about this conceptually, especially when it matters (for a popular posts) you simply cannot trust if tips are real, or people are trolling others attaching non-exiting large transfers.

taggrx: I thought a bit more, if the tx is done on the frontend using user's principal, why not simply checking the sender of the tx to make sure it was actually issued by the user? (we also need to add the post id to the memo and check that so that the user can't attach one tip to many posts). And then if a check is not successful, penalize users by e.g. draining 100 credits, otherwise I would not charge any credits to not penalize tippers.

aligatorr89: We verify tx based on memo which holds postId and index - there is no way for double add. We also check sending principal is the same as from param and amount.

src/backend/env/mod.rs
aligatorr89: I wanted to have tips in stable memory. Could solve it by BTreeMap in stable memory - then this heap index can be removed, wdyt?

src/backend/env/canisters.rs
```diff
@@ -37,6 +38,43 @@ thread_local! {
...
+}
+
+#[derive(
+    Serialize, Deserialize, CandidType, Clone, Hash, Debug, PartialEq, Eq, PartialOrd, Ord, Default,
+)]
+#[serde(transparent)]
+pub struct IcrcMemo(pub ByteBuf);
+
+#[derive(CandidType, Clone, Serialize, Deserialize, Debug)]
+pub struct IcrcTransfer {
```
taggrx: It's the same type as we have in the token.rs module. Let's consolidate?

src/backend/env/tip.rs
```diff
@@ -0,0 +1,384 @@
...
+        .ok_or("sender not found")?;
+    let post = Post::get(state, &post_id).ok_or("post not found")?;
+    let realm = post
+        .realm
+        .as_ref()
+        .and_then(|id| state.realms.get(id))
+        .ok_or("realm not found")?;
+    let memo_u64 = convert_memo_to_u64(memo)?;
+
+    if post.user != receiver_id {
```
taggrx, line 32: Instead of passing both `to_principal` and `post_id` and then making sure all this matches, let's simply pass `post_id` and extract the receiver from it?

src/backend/env/tip.rs
```diff
@@ -0,0 +1,384 @@
...
+        return Err("receiver does not match with post creator".to_string());
+    }
+    if realm
+        .tokens
+        .as_ref()
+        .map_or(false, |tokens| !tokens.contains(&canister_id))
+    {
+        return Err("token is not allowed to tip in realm".to_string());
```
taggrx: Didn't [we agreed](https://6qfxa-ryaaa-aaaai-qbhsq-cai.icp0.io/#/thread/1432672) we don't want to gateway which tokens the user can use to tip?

src/backend/env/canisters.rs
taggrx: But it's not part of any standard agreed by the community for using in canisters. E.g. this wouldn't work for Taggr itself. Why not simply using `icrc3_get_blocks`?

src/backend/env/tip.rs
```diff
@@ -0,0 +1,384 @@
...
+}
+
+#[allow(clippy::too_many_arguments)]
+pub async fn try_tip(
+    post_id: PostId,
+    canister_id: Principal,
+    from_principal: Principal,
+    start_index: u64,
+) -> Result<Tip, String> {
+    // DoS protection
```
taggrx: Please, let's not make tips more unattractive and expensive than they are! Instead let's do this:

- define a local variable with credits penalty for a wrong call (e.g. 20 credits).
- if user has less than needed credits for potential penalty, reject the call.
- fetch the tipping tx
- If the call fails (please don't panic there!) or the tx not found - deduct the penalty.

src/backend/env/tip.rs
```diff
@@ -0,0 +1,384 @@
...
+                    transfer.to.owner,
+                    transfer.from.owner,
+                    start_index,
+                )
+            }),
+            None => Err("Transaction is not a transfer!".into()),
+        }
+    } else {
+        Err(format!(
+            "We could not find transaction at index {}",
```
taggrx: Who is we? :D 

```suggestion
            "Transaction not found at index {}",
```

src/backend/updates.rs
taggrx: > We also check sending principal is the same as from param and amount.

Hm, no. `try_tip` is called with `start_index` that returns a transaction, but we never check if this transaction is in any relation with the principal returned from `read(caller)` or am I missing something?

src/frontend/src/common.tsx
```diff
@@ -472,7 +476,15 @@ export const tokens = (n: number, decimals: number, hideDecimals?: boolean) => {
     });
 };
 
-export const ICP_LEDGER_ID = Principal.fromText("ryjl3-tyaaa-aaaaa-aaaba-cai");
+/** Remove tail zeros */
+export const shortenTokensAmount = (n: number, decimals: number) => {
```
taggrx, line 515: See the `tokens` function in common.tsx, I think this function is not needed.

src/frontend/src/common.tsx
```diff
@@ -1271,3 +1283,250 @@ export const getCanonicalDomain = () =>
...
+                    })
+                    .catch(console.error),
+            ),
+        );
+    }
+
+    return canistersFromStorageMap;
+}
+
+export function numberToUint8Array(num: number) {
```
taggrx, line 1379: Hm, looks like it's not used anywhere?

src/frontend/src/common.tsx
```diff
@@ -1271,3 +1283,250 @@ export const getCanonicalDomain = () =>
...
+
+    // Return the Uint8Array
+    return new Uint8Array(arr.reverse()); // Reverse to ensure correct byte order (little-endian)
+}
+
+type LocalStorageData = {
+    data: any;
+    updated_at: string;
+};
+/** Helper for stale local storage data */
```
taggrx: Let's stick with one style of comments

src/frontend/src/common.tsx
```diff
@@ -1271,3 +1283,250 @@ export const getCanonicalDomain = () =>
...
+    localStorage.setItem(
+        cacheKey,
+        JSON.stringify({
+            data: data,
+            updated_at: new Date().toISOString(),
+        } as LocalStorageData),
+    );
+}
+
+export const Popup = ({
```
taggrx: Let's not introduce another pop-up implementation, there is `popUp` for that (in common).

src/backend/env/canisters.rs
aligatorr89: Taggr tokens defintions are slightly different and is hard to re-use due to integration failures, like `u64` instead of `Nat`. It fails during `deserialize` of response which expects slightly different types.

I think we should rather use ICRC types for Taggr tokens definitions. 

P.S. i re-used Account definition now

src/backend/env/tip.rs
aligatorr89: Receiver is derived from tx retrieved from external ledger

src/backend/env/tip.rs
aligatorr89: We can allow yes, but in the FE they will still see those that realm controller have chosen

src/backend/env/tip.rs
aligatorr89: Makes sense. But we should still charge something for successful tip, how much like 5 ?

src/frontend/src/common.tsx
aligatorr89: It uses tokens, additionally it removes tailing zeros, like `0.020000` -> `0.02`

src/frontend/src/common.tsx
aligatorr89: used in finalizeTip

aligatorr89:
https://github.com/TaggrNetwork/Taggr/pull/87#discussion_r2186904592

It looks interesting, 1 issue is that accounts are as account identifiers, which makes difficult to compare, or is it?

src/backend/env/canisters.rs
aligatorr89: It looks interesting, but `to` and `from` are in some weird format . Any way to convert that?

taggrx: `icrc3_get_blocks` doesn't have any `from` or `to`, I'm not sure what you mean.

src/backend/env/mod.rs
taggrx: Thanks! It's much cleaner and simpler now (and more space efficient!)

src/backend/env/canisters.rs
taggrx: If you'd just switch to the ICRC3 get blocks function, all of this will go away. Let's make progress on the other thread first.

src/backend/env/tip.rs
taggrx: Let's just start with costs of the cheapest reaction?

src/backend/env/tip.rs
taggrx: I see now, thanks! What do you think about passing the entire TX into this function and call it `create_post_tip_from_tx`?

Then:

1. The clippy `too_many_arguments` issue goes away
2. The whole logic is encapsulated in one function and not spread over two funcs.

src/backend/env/tip.rs
taggrx: But I thought the whole point of the discussion was to not have it? See my head message of the thread linked above:

> why a realm should have a list of white listed tokens at all?

src/frontend/src/common.tsx
taggrx: Ah this is nice! Shouldn't we just make this the default display mode of `tokens` function? Just move it there.

src/frontend/src/user_resolve.tsx
```diff
@@ -3,6 +3,10 @@ import { UserId, UserData, User } from "./types";
 import { Loading, commaSeparated, pfpUrl } from "./common";
 
 export const USER_CACHE: UserData = {};
+export const USER_COMMON_CACHE: Record<
+    number,
```
taggrx: Is this the `UserId`?

taggrx: Also, why not simply extending the existing `USER_CACHE` and the type `UserData` with what's needed?

src/frontend/src/wallet.tsx
```diff
@@ -140,7 +140,7 @@ export const Wallet = () => {
                                 ICP_DEFAULT_FEE,
                                 recipient,
                             );
-                            if (typeof response == "string")
+                            if (isNaN(+(response || "")))
```
taggrx:
```suggestion
                            if (Number.isNaN(Number(response || "")))
```

src/frontend/src/wallet.tsx
```diff
@@ -297,8 +297,8 @@ export const Wallet = () => {
                             token_decimals,
                             transaction_fee,
                         );
-                        if (typeof response == "string")
-                            showPopUp("error", response);
+                        if (isNaN(+(response || "")))
```
taggrx:
```suggestion
                        if (Number.isNaN(Number(response || "")))
```

src/frontend/src/common.tsx
aligatorr89: It is used for simple messages. I'm not sure how to make it work for dynamic React HTML - I don't think it's worth trying to re-use it cause we'll just create too many divergencies.

src/frontend/src/user_resolve.tsx
aligatorr89: will try to extend `users_data` to include principal

src/backend/env/tip.rs
aligatorr89: Removed validation

src/backend/env/canisters.rs
aligatorr89: Switching to icrc-3 done :)

src/backend/updates.rs
aligatorr89: added caller check as well

src/backend/env/canisters.rs
aligatorr89: added icrc3_get_blocks

src/frontend/src/common.tsx
aligatorr89: seems a bit difficult, don't want to create more test cases :)

PR #89, opened 2025-04-14
Onboarding via Bitcoin

xqxpx:


cdss3211 (dismissed):
Really want this for myself but I'm too ignorant and really just lost on what needs to happen to make things work out for the better for all parties involved. So if there's a brave enough soul on God's precious earth that will contact me directly and shoot me straight without trying to get over on me I promise you I will protect and put every ounce of energy within me to make sure that our relationship's/friendship work. That's my word, and I stand on that. You can't blame a man for whatever if a man doesn't even know what he has? But now it's starting to hit home. So hey someone please reach out to me personally please I'm pleading with whoever.

src/backend/env/invoices.rs
```diff
@@ -66,19 +91,57 @@ impl Invoices {
...
 
-    pub fn close(&mut self, invoice_id: &Principal) {
-        self.invoices.remove(invoice_id);
+    async fn new_btc_invoice(invoice_id: Principal, sats: u64) -> Result<BTCInvoice, String> {
+        if sats == 0 {
+            return Err("wrong USD/BTC ratio".into());
+        }
+        // The derivation path contains the timestamp and the principal.
+        let derivation_path = vec![
+            time().to_be_bytes().to_vec(),
```
6qfxa, line 100: Are you adding the timestamp to make the derivation path unique? If so, then this is not guaranteed.

A user can make multiple calls to `mint_credits_with_btc()` at the same time. If they run in the same round, then `time()` will be the same for them.

src/backend/env/invoices.rs
```diff
@@ -129,6 +192,145 @@ impl Invoices {
...
+        if invoice.tx_id.as_ref().is_some() {
+            let result = bitcoin::balance(invoice.address.clone()).await;
+            match result {
+                Ok(balance) => {
+                    if balance < invoice.balance {
+                        total_sats += invoice.balance;
+                        continue;
+                    } else {
+                        // Increase fees by two percentiles.
+                        invoice.fee_percentile = (invoice.fee_percentile + 2) % 100;
```
6qfxa, line 275: Could this cause paying more for the transfer than the balance of the invoice?

6qfxa (commented):
I just skimmed the PR. It looks reasonable AFAICS.

General comments:
- maybe add canister balance checks before making the expensive calls. If the balance is below some threshold, then log an error and bail out to avoid draining the cycles to 0.
- maybe disallow the same principal from creating multiple concurrent btc invoices.

src/backend/env/invoices.rs
taggrx: But wouldn't we overwrite the first invoice (which is fully ok) because we're using the principal id as lookup key?

src/backend/env/invoices.rs
taggrx: Absolutely yes. But if we assume that the primary goal of this is spam protection and creating a bridge into the BTC world and it allows us to keep the implementation very simple, I personally think it's fine. WDYT?

PR #106, opened 2025-06-04
feat: adds toggle for displaying previous usernames

krpeacock:
image

For users who have changed account handles, adds a toggle to display or hide previous account names on their public profile. Doesn't change the record or forwarding, just the display of the AKA field. Defaults to true

Developer Note: leaving the field in but hidden is intentional - this way the value gets set during profile updates or creation.

### DIFF
--- a/src/frontend/src/settings.tsx
+++ b/src/frontend/src/settings.tsx
@@ -168,6 +168,22 @@ export const Settings = ({ invite }: { invite?: string }) => {
         await window.reloadUser();
     };
 
+    const handleCheckboxSetting = (
+        key: string,
+        event: React.ChangeEvent<HTMLInputElement>,
+    ) => {
+        const { checked } = event.target;
+        switch (checked) {
+            case true:
+                event.target.value = "on";
+                break;
+            case false:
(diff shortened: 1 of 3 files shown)


src/frontend/src/settings.tsx
```diff
@@ -183,6 +199,26 @@ export const Settings = ({ invite }: { invite?: string }) => {
                     placeholder="alphanumeric"
                     onChange={namePicker}
                 />
+                {
+                    <div
+                        className={`bottom_spaced ${user?.previous_names?.length > 0 ? "" : "hidden"}`}
```
taggrx, line 204: I think it's better to simply condition the whole DOM element on the length of prev names instead of doing it via styling, e.g.:

```js
{user.previous_names.length > 0 && <div> ... </div>}
```

xqxpx (commented):
Thanks @krpeacock!

Quick questions: could you please explain the "itch" you're "scratching" here, namely, what's the prupose of previously used names on the settings screen?

xqxpx:
Closing for now due to stale discussion.

PR #121, opened 2025-10-28
Use get_transactions instead of icrc3_get_blocks

aligatorr89:
Use get_transactions instead of icrc3_get_blocks since it's supported widely


xqxpx:
Merged this into the release branch

